Issue #1, opened 2018-05-15
Organize the .yaml file


fiennyangeln:
Organize the .yaml files (one .yaml file per repository) and put them in an appropriate directory

djmitche:
It looks like this is done:
  https://github.com/mozilla-frontend-infra/bug-issue-ui/tree/master/src/data

can it be closed?

fiennyangeln:
@djmitche sure!

PR #4, opened 2018-05-16
Add initial yaml files

fiennyangeln:
- Each super group has its own yaml file
- Each yaml file contain the projects (group) based on each category (supergroup)
  e.g : Support (supergroup) has tascluster, releng, automation as its group
- Each group have name, which will appear as an option in the sidebar
- Each group also has info header and subinfo (I named it subinfo since we might want to add the definition of project as info)
- Each group has product which may / may not have components (if it has it will have format of [product] : [list of component] , otherwise the format is [product])
-  Each group can also have Repositories with the format [Repo]: [Tag]

Bug #1


src/data/firefox.yaml
```diff
@@ -0,0 +1,65 @@
+Firefox:
+  ff:
```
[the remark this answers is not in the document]
fiennyangeln: then the file is located on the folder named the super group (so in total 5 folder)?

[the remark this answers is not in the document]
eliperelman: I'm not convinced we need the super group at all. What information does a supposed super group provide that isn't given by each group?

fiennyangeln:
@eliperelman I modified it, please review:D

src/App.jsx
```diff
@@ -104,7 +106,8 @@ class App extends Component {
 
   render() {
     const { classes, theme } = this.props;
-    const { projectSelected, allProjects } = this.state;
+    const { allProjects } = this.state;
```
helfi92: What do you think about renaming `allProjects` to `projects`.

src/App.jsx
```diff
@@ -90,6 +91,7 @@ class App extends Component {
   };
 
   handleCheckboxToggle = (proj) => {
+
```
helfi92: Remove newline.

src/App.jsx
```diff
@@ -20,6 +20,7 @@ import Hidden from 'material-ui/Hidden';
 import IconButton from '@material-ui/core/IconButton';
 import MenuIcon from '@material-ui/icons/Menu';
 
+
```
helfi92: If we are putting local imports at the end, then we don't necessarily need to have added newlines here.

src/App.jsx
fiennyangeln: okay! it is inexistent in the next PR but should I need it again I will change it to `projects`

fiennyangeln:
@helfi92 ok modified!:D

src/data/Web Development/bugzilla.yaml
```diff
@@ -0,0 +1,10 @@
+name: Bugzilla
+description: |
+  ### Extra information for Bugzilla
+
+  
+
+  #### Get involved with the [Bugzilla](https://wiki.mozilla.org/Bugzilla) project or [BMO](https://wiki.mozilla.org/BMO), Mozilla's custom version
```
helfi92: Could we remove the `####`. As a result, I think the markdown component will wrap it in a  `p` tag by default.

```
### Extra information for Bugzilla

Get involved with the [Bugzilla](https://wiki.mozilla.org/Bugzilla) project or [BMO](https://wiki.mozilla.org/BMO), Mozilla's custom version
```

If you could do the same thing for the other `yaml` files, that would be awesome :)

fiennyangeln: @helfi92 Sure, I modified it. Please check :D

helfi92: @fiennyangeln Any reason why 3 new lines between `###` and the description? Why not 1?

fiennyangeln: It is because the library that I use html2text add new lines in between the tag and initially I think it has to be that way.. Changed :D thanks for the catch

eliperelman:
@fiennyangeln are you able to re-flag us for review when you need it; does the GitHub UI let you do that?

src/App.jsx
```diff
@@ -74,13 +74,13 @@ class App extends Component {
 
   constructor(props) {
     super(props);
-    var allProjects = [...data.Projs];
-    allProjects.forEach(function(project, index) {
+    var projects = [...data.Projs];
+    projects.forEach(function(project, index) {
```
eliperelman: Prefer arrow functions, otherwise you won't be able to access `this` from the class without a re-bind.

src/App.jsx
```diff
@@ -90,29 +90,30 @@ class App extends Component {
   };
 
   handleCheckboxToggle = (proj) => {
-    var allProjects = [...this.state.allProjects];
 
-    allProjects.forEach(function(project, index) {
+    var projects = [...this.state.projects];
+    projects.forEach(function(project, index) {
```
eliperelman: Use arrow functions.

src/App.jsx
```diff
@@ -90,29 +90,30 @@ class App extends Component {
...
       if (project[0] === proj) {
         console.log(project[2])
```
eliperelman: Remove extra `console.log`.

src/data/Firefox/devtools.yaml
```diff
@@ -0,0 +1,6 @@
+name: Developer Tools
```
eliperelman: I'm still not sure why we need the extra folders. Can we put all of these inside the `data` folder?

src/data/Web Development/bugzilla.yaml
```diff
@@ -0,0 +1,7 @@
+name: Bugzilla
+description: |
+  ### Extra information for Bugzilla
```
eliperelman: I don't think we need these `Extra information` headings.

src/data/Firefox/devtools.yaml
fiennyangeln: because I need to keep track of the group each project belongs to.. do u think its better if i add it as a field in the yaml file?

src/data/Web Development/bugzilla.yaml
helfi92: Yeah.

src/data/Firefox/devtools.yaml
eliperelman: That's the crux of the issue I am getting at: do we even *need* to track what group it belongs to? My instinct says no.

fiennyangeln: @eliperelman  in the current bugsahoy, the project is listed based on its group (e.g Firefox with its User Interface, Developer Tools, Internal, Internal (iOS), etc), or do you suggest changing the name to be Firefox : User Interface, Firefox: Developer Tools ? so we can place all of them in the same directory

src/App.jsx
fiennyangeln: Noted. changed in other PR! Thx!!:D

fiennyangeln:
@eliperelman i dont think re-flag is possible, btw, I modified the description.

PR #6, opened 2018-05-16
Display project based on yaml file

fiennyangeln:
- modify the preset from @neutrinojs/react to neutrino-preset-mozilla-frontend-infra
- add lint script in package.json
- add loader.js which will read from yaml files located in data
- modify index.jsx according to linter standard
- modify app.jsx to display the project list (one project per line) based on data from loader.js


src/loader.js
```diff
@@ -0,0 +1,15 @@
+import Firefox from './data/firefox.yaml';
```
eliperelman: As the list of yml files grows, so will having to maintain this file. You can use something like the following to do it dynamically:

```js
const importer = require.context('./', true, /\.yml/);
const keys = [...new Set([...importer.keys()])];

module.exports = keys.reduce(
  (resolvers, key) => ({ ...resolvers, [key]: importer(key).default }),
  {}
);
```

helfi92: Love this.

helfi92: The [website](http://yaml.org/faq.html) seems to recommend using `yaml` instead of `yml` for the extension.

helfi92: We can perhaps add a `$` symbol to the end of the regex to make sure it doesn't match with something like `yaml.js`:

```js
const importer = require.context('./', true, /\.yaml$/);
```

eliperelman: Weird, I don't know any project that uses yaml over yml, like Travis, Taskcluster, etc. That said we could support both with `/\.(yml|yaml)$/`.

src/data/webplatform.yaml
```diff
@@ -0,0 +1,41 @@
...
+    Name: DOM and CSS technology
+    Products: 
+     - Core: ['CSS Parsing and Computation', 'SVG', 'DOM', 'DOM: Core & HTML', 'DOM: CSS Object Model', 'DOM: Device Interfaces', 'DOM: Events', 'DOM: IndexedDB', 'DOM: Mozilla Extensions', 'DOM: Other', 'DOM: Traversal-Range', 'DOM: Validation', 'DOM: Workers', 'Geolocation', 'HTML: Form Submission', 'Event Handling', 'HTML: Parser', 'MathML', 'XML', 'XSLT']
+  editor:
+    Name: Input handling
+    Products: 
+     - Core: ['Editor', 'Selection', 'Keyboard: Navigation', 'Drag and Drop', 'Spelling Checker']
+  media:
+    Name: Media
+    Products:
```
helfi92: The file is becoming lengthy. Maybe we can separate subprojects into their own yaml files.

src/App.jsx
```diff
@@ -65,100 +58,127 @@ const styles = theme => ({
...
-                  control={
-                    <Checkbox
-                      value={bug}
-                    />
-                  }
-                  label={bug}
-                  key={index}
-                />
+        <FormControl component="fieldset">
+          <FormLabel component="legend">Are you interested in:</FormLabel>
```
helfi92: I think `FormControl` is usually used for form inputs. The drawer can be done using plain `div`s. This is how we've done it in tc-web https://github.com/taskcluster/taskcluster-web/blob/4e75cb216cdede3fc40e75837e6b7b94f180c761/src/components/Dashboard/index.jsx#L163-L187.

src/App.jsx
```diff
@@ -65,100 +58,127 @@ const styles = theme => ({
...
+        const [key, proj] = project;
+
+        proj.key = key;
+        projects.push(proj);
+        projectSelected[key] = false;
+
+        return projects;
+      }, []);
+      allProjectGroupsEntries.push(curr);
+    });
```
helfi92: If we have complete control of the yaml files, could we organize them in a way that wouldn't require organizing them after loading them?

src/data/webplatform.yaml
fiennyangeln: @helfi92 yes I think of that too but initially i am a bit confused of way to import bunch of files. I modified it in other PR.

src/App.jsx
fiennyangeln: Yeah I really hope it could be that way too. The problem I face so far is when it is recognized as key value pair and I need both of them.. 

`Name: Accessibility
InfoHeader: Extra information for Accessibility
SubInfo: Get involved with the [Accessibility](https://wiki.mozilla.org/Accessibility/Contribute) team
Products: 
 - Core: ['Disability Access APIs']
 - Firefox: ['Disability Access']
 - additional: {'keywords': ['access']}`

in the above, previously, we load the file and import it using the file name 'a11y' as the key and the above as the value. But then I realized it might be better to keep the 'a11y' as well, to later be the value of the checkbox and use 'Accessibility' as the displayed name only as 'Accessibility' might have duplicate in other SuperGroup  or i might need 'a11y' as the query later..

I initially think of adding 'Key: a11y' but i am afraid it will be redundant and I think later we might face similar problem when handling the Products (if we need both the 'Core' and ['Disability Access APIs'] )

src/App.jsx
```diff
@@ -65,100 +53,107 @@ const styles = theme => ({
...
-            </FormGroup>
+        <FormLabel component="legend">Are you interested in:</FormLabel>
 
-            <FormLabel component="legend">Do you know ?</FormLabel>
-            <FormGroup>
-              { data.Languages.map((lang, index) => (              
+        <FormGroup>
+          {Object.entries(projectGroups).map(([group, projects]) => (
+            <div key={group}>
+              <Typography variant="subheading">{group}</Typography>
```
helfi92: Can we make use of the `List` component here? This would add styling for list items like paddings and we would be able to add the on hover effect to when a user hovers over the items.

```js
{Object.entries(projectGroups).map(([group, projects]) => (
  <List>
    <ListItem>
      <ListItemText>{group}</ListItemText>
    </ListItem>
    {projects.map(project => (
      <ListItem key={`${group}-${project.fileName}`} id={project.fileName} onClick={this.handleCheckboxToggle} button>
        <Checkbox
          value={project.fileName}
          checked={isProjectSelected[project.fileName]}
        />
        <ListItemText primary={project.name} />
      </ListItem>
    ))}
  </List>
))}
```

Notice that I also changed

```js
onChange={e =>this.handleCheckboxToggle(project.fileName, e)}
```
to:

```js
id={project.fileName} onClick={this.handleCheckboxToggle}
```

Because of that, you will need to change `handleCheckboxToggle` to extract the value from the id:

```js
handleCheckboxToggle = ({ currentTarget: { id } }) => {
  const { isProjectSelected } = this.state;

  isProjectSelected[id] = !isProjectSelected[id];
  this.setState({ isProjectSelected });
};
```

PR #15, opened 2018-05-22
Create searchbar

fiennyangeln:
Create search bar to filter the view based on project acronym (filename) or project name


src/components/CustomizedCard/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+    padding: '15px 20px',
+    margin: '0 0 10px',
+  },
+  cardActions: {
+    height: 'auto',
+    display: 'inline',
+  },
+});
+
+function CustomizedCard({ ...props }) {
```
helfi92: This component shows information about a project. Could we rename it to `ProjectCard`?

src/components/CustomizedCard/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+          </Typography>
+        ) : null}
+        {description !== undefined ? (
+          <Typography className={classes.cardDescription}>
+            {description}
+          </Typography>
+        ) : null}
+      </CardContent>
+      <CardActions className={`${classes.textAlign} ${classes.cardActions}`}>
+        {footer}
```
helfi92: Can we use `CardActions` here for the footer? You will probably not need the classes if you use that component.

src/views/Projects/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+      .reduce(
+        (prev, key) => ({
+          ...prev,
+          [key]: projects[key],
+        }),
+        {}
+      );
+
+    return (
+      <div className={classes.root}>
```
helfi92: The styling of `classes.root` is not needed. See https://github.com/mozilla-frontend-infra/bug-issue-ui/pull/12#discussion_r189601958 for how I was able to go remove it.

src/components/CustomizedCard/index.jsx
helfi92: https://material-ui.com/demos/cards/#simple-card

src/views/Projects/index.jsx
```diff
@@ -0,0 +1,147 @@
...
+            spacing={16}
+            direction="column"
+            justify="center"
+            alignItems="center">
+            <Grid item xs={12}>
+              <div className={classes.columnFlex}>
+                <div className={classes.flex1} />
+                <div>
+                  <Typography variant="display4" align="center">
+                    Bug Issue UI
```
helfi92: I think you can just use Bug Issue.

helfi92: I don't think you need of all those `div`s wrappers. See https://github.com/mozilla-frontend-infra/bug-issue-ui/pull/12#discussion_r189601958 for how you can simply put it under a `Fragment`. Since the title will always be on its own line, we also don't need it to be wrapped inside a `Grid`.

src/components/CustomizedCard/index.jsx
```diff
@@ -0,0 +1,96 @@
+import {
+  withStyles,
+  Card,
+  CardContent,
+  CardActions,
+  Typography,
+} from 'material-ui';
+import PropTypes from 'prop-types';
```
helfi92: Only import what you need `import { node, string, ... } from 'prop-types';`.

src/components/CustomizedCard/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+CustomizedCard.propTypes = {
+  classes: PropTypes.object.isRequired,
+  title: PropTypes.node,
+  subtitle: PropTypes.node,
+  description: PropTypes.node,
+  footer: PropTypes.node,
+  avatar: PropTypes.string,
+};
```
helfi92: For properties that are not required, we would need to set defaults for them:

```js
CustomizeCard.propTypes = {
  /** Override or extend the styles applied to the component. */
  classes: object,
  /** The project title. */
  title: string.isRequired,
  /** The project subtitle. */
  subtitle: string,
  /** The project description in markdown. */
  description: string.isRequired,
}
```

```js
CustomizeCard.defaultProps = {
  classes: null,
  subtitle: null,
};
```

src/components/CustomizedCard/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+          </Typography>
+        ) : null}
+        {title !== undefined ? (
+          <Typography component="h4" className={classes.cardTitle}>
+            {title}
+          </Typography>
+        ) : null}
+        {description !== undefined ? (
+          <Typography className={classes.cardDescription}>
+            {description}
```
helfi92: `<ReactMarkdown source={description} />`

src/components/CustomizedCard/index.jsx
fiennyangeln: @helfi92 I think I used CardActions, I use the classes to make the text centered

fiennyangeln:
@helfi92 I think all of the review are for the homepage so I push the change in the other PR. Should I do rebase into here?

fiennyangeln:
@helfi92 Yeah I think I will create a new PR if we `git reset --soft HEAD^^ git commit --amend` later and rebasing takes too much time.

Issue #14, opened 2018-05-22
Create search bar


fiennyangeln:
Create search bar to filter based on projects

PR #22, opened 2018-05-31
Create a search bar in the homepage

fiennyangeln:
Search bar will enable filtering on Project Name based on user input

Bug #14

### DIFF
--- a/src/views/Projects/index.jsx
+++ b/src/views/Projects/index.jsx
@@ -2,6 +2,7 @@ import { Component, Fragment } from 'react';
 import { withStyles } from 'material-ui/styles';
 import Grid from 'material-ui/Grid';
 import Typography from 'material-ui/Typography';
+import TextField from 'material-ui/TextField';
 import projects from '../../data/loader';
 import ProjectCard from '../../components/ProjectCard';
 
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
 export default class Projects extends Component {
   state = {
     projects,
+    search: '',
+  };
+  handleTextInputChange = event => {
+    this.setState({ search: event.target.value });
   };
 
   render() {
     const { classes } = this.props;
-    const { projects } = this.state;
+    const { projects, search } = this.state;
+    const selectedProjects = Object.keys(projects)
+      .filter(fileName => {
+        const found =
+          fileName.toLowerCase().indexOf(search.toLowerCase()) !== -1 ||
+          projects[fileName].name
+            .toLowerCase()
+            .indexOf(search.toLowerCase()) !== -1;
+
+        return found;
+      })
+      .reduce(
+        (prev, key) => ({
+          ...prev,
+          [key]: projects[key],
+        }),
+        {}
+      );
 
     return (
       <Fragment>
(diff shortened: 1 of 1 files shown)


src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
 export default class Projects extends Component {
   state = {
     projects,
```
helfi92: `projects` doesn't need to be in state because you are not doing `this.setState({ projects: ... })` anywhere in the component.

src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
...
+    search: '',
+  };
+  handleTextInputChange = event => {
+    this.setState({ search: event.target.value });
   };
 
   render() {
     const { classes } = this.props;
-    const { projects } = this.state;
+    const { projects, search } = this.state;
```
helfi92: Remove `projects` from state. Replace `search` with `searchTerm`.

src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
...
+    const selectedProjects = Object.keys(projects)
+      .filter(fileName => {
+        const found =
+          fileName.toLowerCase().indexOf(search.toLowerCase()) !== -1 ||
```
helfi92: A user, in theory, has no knowledge of filenames so we don't need to filter on them. Filtering on project names is good enough.

src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
...
+          projects[fileName].name
+            .toLowerCase()
+            .indexOf(search.toLowerCase()) !== -1;
```
helfi92: You can make usage of `Array.prototype.includes` instead of `indexOf`. Less code to write and more clarity.

src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
...
+  };
+  handleTextInputChange = event => {
+    this.setState({ search: event.target.value });
   };
 
   render() {
     const { classes } = this.props;
-    const { projects } = this.state;
+    const { projects, search } = this.state;
+    const selectedProjects = Object.keys(projects)
```
helfi92: We are filtering on projects not necessarily selecting. I think something like `filteredProjects` is clearer.

src/views/Projects/index.jsx
```diff
@@ -20,11 +21,32 @@ import ProjectCard from '../../components/ProjectCard';
...
+
+        return found;
+      })
+      .reduce(
+        (prev, key) => ({
+          ...prev,
+          [key]: projects[key],
+        }),
+        {}
+      );
```
helfi92: We can replace all that logic with:

```js
const filteredProjects = Object.keys(projects)
  .filter(fileName =>
    projects[fileName].name.toLowerCase().includes(searchTerm.toLowerCase())
  )
  .reduce(
    (prev, key) => ({
      ...prev,
      [key]: projects[key],
    }),
    {}
  );
```

src/views/Projects/index.jsx
```diff
@@ -37,8 +59,17 @@ export default class Projects extends Component {
           </Typography>
         </header>
         <main className={classes.container}>
+          <TextField
+            label="Search"
+            id="margin-normal"
```
helfi92: Remove `id`.

src/views/Projects/index.jsx
```diff
@@ -37,8 +59,17 @@ export default class Projects extends Component {
...
+            fullWidth
+            value={this.state.search}
+            className={classes.textField}
+            helperText="Project name, Keyword"
```
helfi92: For a simple field like this, we won't need a "helper" text. Remove this line.

src/views/Projects/index.jsx
fiennyangeln: Ok! wah!! i didnt know it could be this simple 💯

PR #29, opened 2018-06-07
Change the site name

fiennyangeln:
Change the site name from bug-issue to codetribute

Bug #28

### DIFF
--- a/src/views/Projects/index.jsx
+++ b/src/views/Projects/index.jsx
@@ -51,10 +51,10 @@ export default class Projects extends Component {
       <Fragment>
         <header className={classes.header}>
           <Typography variant="display2" align="center">
-            Bug Issue
+            Codetribute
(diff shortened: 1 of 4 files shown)


src/views/Projects/index.jsx
```diff
@@ -51,10 +51,10 @@ export default class Projects extends Component {
       <Fragment>
         <header className={classes.header}>
           <Typography variant="display2" align="center">
-            Bug Issue
+            Codetribute
           </Typography>
           <Typography variant="subheading" align="center">
-            Find your first contribution with Mozilla
+            Code to contribute with Mozilla
```
helfi92: What do you think about "Find your first code contribution with Mozilla".

fiennyangeln: Sure!

PR #31, opened 2018-06-07
Add link for each bug/issue entry in the table

fiennyangeln:
Bug #26

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -20,6 +20,9 @@ const sorted = pipe(
   tableWrapper: {
     overflowX: 'auto',
   },
+  link: {
+    textDecoration: 'none',
+  },
 }))
 export default class TasksTable extends Component {
   static propTypes = {
@@ -76,7 +79,15 @@ export default class TasksTable extends Component {
         <DataTable
           items={data}
           renderRow={item => (
-            <TableRow hover tabIndex={-1} key={`${item.summary}`}>
+            <TableRow
+              hover
+              target="_blank"
+              rel="noopener noreferrer"
+              tabIndex={-1}
+              key={item.summary}
+              component="a"
+              href={item.url}
+              className={classes.link}>
               <TableCell component="th" scope="row">
                 {item.project}
               </TableCell>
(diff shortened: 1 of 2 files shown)


src/components/TasksTable/index.jsx
```diff
@@ -76,7 +79,13 @@ export default class TasksTable extends Component {
...
           items={data}
           renderRow={item => (
-            <TableRow hover tabIndex={-1} key={`${item.summary}`}>
+            <TableRow
+              hover
+              tabIndex={-1}
+              key={`${item.summary}`}
+              component="a"
+              href={item.url}
+              className={classes.link}>
```
helfi92: When we are dealing with an external link, the convention is navigate the user to the new page via a new window. In code, we do that with `target="_blank"`. Also, to make sure the new window open doesn't have access to the window who opened it (tabnabbing), we add `rel="noopener noreferrer"`. For more information on this, you can refer to this [article](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/).

```jsx
<TableRow
  target="_blank"
  rel="noopener noreferrer"
  // ...
/>
```

src/components/TasksTable/index.jsx
```diff
@@ -76,7 +79,13 @@ export default class TasksTable extends Component {
         <DataTable
           items={data}
           renderRow={item => (
-            <TableRow hover tabIndex={-1} key={`${item.summary}`}>
+            <TableRow
+              hover
+              tabIndex={-1}
+              key={`${item.summary}`}
```
eliperelman: Template strings are unnecessary when the content of the interpolation is just a single string:

```jsx
key={item.summary}
```

src/views/Project/index.jsx
```diff
@@ -11,6 +11,7 @@ const data = [
     tag: 'JS',
     assignee: 'None',
     lastUpdated: '2018-05-20',
+    url: 'https://bugzilla.mozilla.org/show_bug.cgi?id=1436212',
```
eliperelman: Note sure why we are adding these links when I believe this whole content is being removed in a separate PR.

fiennyangeln: @eliperelman i think we need a link in the table entry so that user can click on the entry and get redirect to the bug page, be it in github or bugzilla.

helfi92: Having something hardcoded for now will be good for demoing the site during the all-hands.

Issue #34, opened 2018-06-08
Enhance the Last Updated column in the project table


helfi92:
The last updated column in the project table shows the date in the long format (2018-05-04T21:58:21Z).  It would be nice if we can have last updated to be something like "2 days ago". We can achieve that with https://date-fns.org/v2.0.0-alpha.9/docs/formatDistance.

PR #37, opened 2018-06-09
Use distance to display the lastUpdated column in table

fiennyangeln:
Bug #34


src/components/TasksTable/index.jsx
```diff
@@ -66,6 +67,8 @@ export default class TasksTable extends Component {
     this.setState({ sortBy, sortDirection });
   };
 
+  renderDate = date => `${formatDistance(date, new Date())} ago`;
```
helfi92: Use the `addSuffix` option instead of `ago`. https://date-fns.org/v2.0.0-alpha.9/docs/formatDistance#arguments

helfi92: Inline the jsx in the render method.

Issue #21, opened 2018-05-28
Fetch data from Bugzilla using Graphql


fiennyangeln:
using the bugzilla graphql gateway repository inside mozilla frontend infra

PR #42, opened 2018-06-14
Fetch bugs from bugzilla and display it in the table

fiennyangeln:
Fetch bugs from bugzilla and display it in the table

The initial fetch in the decorator is either :
1. Product without component ( which can be aggregated together)
if it does not exist ..
2. First pair of products and its component

Bug #21


src/views/Project/index.jsx
```diff
@@ -79,22 +152,44 @@ export default class Project extends Component {
...
-        return this.fetch(searchQuery);
+        return this.fetchGithub(searchQuery);
+      })
+    );
+    const variable = {
+      search: {
+        tags: ['good-first-bug'],
+        statuses: ['NEW', 'UNCONFIRMED', 'ASSIGNED', 'REOPENED'],
+      },
+      paging: {
```
helfi92: The `paging` field will never change so we should probably just add it directly in `fetchMore`.

src/views/Project/index.jsx
```diff
@@ -79,22 +152,44 @@ export default class Project extends Component {
...
         ].join(' ');
 
-        return this.fetch(searchQuery);
+        return this.fetchGithub(searchQuery);
+      })
+    );
+    const variable = {
+      search: {
+        tags: ['good-first-bug'],
+        statuses: ['NEW', 'UNCONFIRMED', 'ASSIGNED', 'REOPENED'],
```
helfi92: The `tags` and `statuses` field doesn't seem to vary so we should probably just add it directly in fetchMore.

src/views/Project/index.jsx
```diff
@@ -79,22 +152,44 @@ export default class Project extends Component {
...
+    const productWithComponentList = mergeAll(
+      project.products.filter(product => typeof product !== 'string') || []
+    );
+
+    await Promise.all(
+      Object.entries(productWithComponentList).map(([products, components]) => {
+        variable.search.products = [products];
+        variable.search.components = components;
+
+        return this.fetchBugzilla(variable);
```
helfi92:
```jsx
Object.entries(productWithComponentList)
  .map(([products, components]) => this.fetchBugzilla([products], components));
```

src/views/Project/index.jsx
```diff
@@ -40,16 +79,50 @@ export default class Project extends Component {
     this.load();
   }
 
-  fetch = searchQuery => {
+  fetchBugzilla = variable => {
```
helfi92: `fetchBugzilla = (products, components) => {`.

src/views/Project/index.jsx
```diff
@@ -79,22 +152,44 @@ export default class Project extends Component {
...
+        tags: ['good-first-bug'],
+        statuses: ['NEW', 'UNCONFIRMED', 'ASSIGNED', 'REOPENED'],
+      },
+      paging: {
+        page: 0,
+        pageSize: 100,
+      },
+    };
+    const productWithComponentList = mergeAll(
+      project.products.filter(product => typeof product !== 'string') || []
```
helfi92: Running the app locally and opening and opening `http://localhost:5000/projects/servo` I get `project.products is undefined` in the console with an infinite spinner.

src/views/Project/index.jsx
```diff
@@ -22,15 +23,53 @@ const tagReposMapping = repositories =>
...
+  graphql(bugsQuery, {
+    skip: props => !projects[props.match.params.project].products,
+    name: 'bugzilla',
+    options: props => ({
+      variables: {
+        search: {
+          products:
+            projects[props.match.params.project].products.filter(
+              product => typeof product === 'string'
+            ).length > 0
```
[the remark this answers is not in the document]
fiennyangeln: @helfi92 typeof product is string if it does not contain component ( - Taskcluster), and it is an object if it contains component ( - Taskcluster: ['Queue' ,'Tools'])

src/views/Project/index.jsx
fiennyangeln: ok fixed

fiennyangeln:
@helfi92 you can try click the addons.mozilla.org or test automation, but you also need to run the bugzilla-graphql-gateway :)

src/App.jsx
```diff
@@ -17,14 +18,19 @@ export default class App extends Component {
     cache: this.cache,
     storage,
   });
-  apolloClient = new ApolloClient({
-    cache: this.cache,
-    link: new HttpLink({
+  link = new RetryLink().split(
```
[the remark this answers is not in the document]
fiennyangeln: I think it is to retry the fetch in case of network error.

src/views/Project/index.jsx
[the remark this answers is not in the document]
helfi92: I see. In that case, we can create a function `componentsFromProducts` which takes in a product object and returns an array of components. The latter can return an empty array if the product has no components. To check for emptiness, we can do `!componentsFromProducts(products).length`

As a result logic duplication is removed and the code becomes more readable.

src/views/Project/index.jsx
```diff
@@ -22,15 +23,58 @@ const tagReposMapping = repositories =>
...
+    skip: props => !projects[props.match.params.project].products,
+    name: 'bugzilla',
+    options: props => ({
+      variables: {
+        search: {
+          tags: ['good-first-bug'],
+          products:
+            projects[props.match.params.project].products.filter(
+              product => typeof product === 'string'
+            ).length > 0
```
helfi92: We can create a function `componentsFromProducts` which takes in a product object and returns an array of components. The latter can return an empty array if the product has no components. To check for emptiness, we can do `!componentsFromProducts(products).length`

As a result logic duplication is removed and the code becomes more readable.

src/App.jsx
```diff
@@ -4,6 +4,7 @@ import { BrowserRouter, Switch, Route } from 'react-router-dom';
 import { ApolloProvider } from 'react-apollo';
 import { ApolloClient } from 'apollo-client';
 import { HttpLink } from 'apollo-link-http';
+import { RetryLink } from 'apollo-link-retry';
 import { InMemoryCache } from 'apollo-cache-inmemory';
 import { CachePersistor } from 'apollo-cache-persist';
```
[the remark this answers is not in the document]
fiennyangeln: I thought it is used below with default arguments to split two links based on context?

fiennyangeln:
I realized the reason why the bug is not showing is because I added tags instead of keywords, and also use the projects.products instead of project.products in the productWithComponentList, sorry!

src/App.jsx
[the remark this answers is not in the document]
helfi92: It's assigned to `persistence` but that variable is unused. We can just do:

```jsx
new CachePersistor({
  cache: this.cache,
  storage,
});
```

src/views/Project/index.jsx
```diff
@@ -12,6 +12,16 @@ import TasksTable from '../../components/TasksTable';
 import issuesQuery from './issues.graphql';
 import bugsQuery from './bugs.graphql';
 
+const componentsFromProduct = product => {
```
helfi92: Can we change this to `componentsFromProducts` and accept a list of products instead of 1 product? This would remove the need to have many `.filter` calls in the file.

helfi92: Sorry, I was thinking something like:

```jsx
const productsList = (products) =>
  products.filter(product => typeof product === 'string');


// search query
products:
  productsList(projects[project].products).length
    ? productsList(projects[project].products)
    : Object.keys(projects[project].products[0])[0],
components: ...
```
Will this work?

fiennyangeln:
@helfi92 ok changed! i move the cache outside the component, and use the persistCache instead https://github.com/apollographql/apollo-cache-persist

src/views/Project/index.jsx
```diff
@@ -55,23 +47,19 @@ const tagReposMapping = repositories =>
...
           keywords: ['good-first-bug'],
           products:
-            projects[props.match.params.project].products.filter(
-              product => !componentsFromProduct(product).length
-            ).length > 0
-              ? projects[props.match.params.project].products.filter(
-                  product => !componentsFromProduct(product).length
-                )
+            productsList(projects[props.match.params.project].products).length >
+            0
```
helfi92: When checking for elements in an array, instead of doing `.length > 0`, you can simple do `.length`:

```jsx
productsList(projects[project].products).length ?
  ... :
  ...;
```

This will return true when length > 0. False otherwise.

Next, instead of using `props.match.params.project` anytime we want to access `project`, you can do:

```jsx
options: ({ match: { params: { project } } }) => ({
```

and just use `project` instead.

src/views/Project/index.jsx
```diff
@@ -22,15 +25,58 @@ const tagReposMapping = repositories =>
...
+    }) => !projects[project].products,
+    name: 'bugzilla',
+    options: ({
+      match: {
+        params: { project },
+      },
+    }) => ({
+      variables: {
+        search: {
+          keywords: ['good-first-bug'],
```
helfi92: `keywords` and `statuses` can be extracted to the top and reused here and inside `fetchBugzilla` via the spread operator:

```jsx
const bugzillaSearchOptions = {
  keywords: ['good-first-bug'],
  statuses: ['NEW', 'UNCONFIRMED', 'ASSIGNED', 'REOPENED'], 
};
```

src/views/Project/index.jsx
```diff
@@ -40,16 +86,61 @@ export default class Project extends Component {
...
     const {
-      data: { fetchMore },
+      bugzilla: { fetchMore },
+    } = this.props;
+
+    return fetchMore({
+      query: bugsQuery,
+      variables: {
+        search: {
+          keywords: ['good-first-bug'],
```
helfi92: Remove `keywords` and `statuses` and use `...bugzillaSearchOptions,` instead.

src/views/Project/index.jsx
```diff
@@ -22,15 +25,58 @@ const tagReposMapping = repositories =>
...
+          products: productsList(projects[project].products).length
+            ? productsList(projects[project].products)
+            : Object.keys(projects[project].products[0])[0],
+          components: productsList(projects[project].products).length
+            ? undefined
+            : Object.values(projects[project].products[0])[0],
+          statuses: ['NEW', 'UNCONFIRMED', 'ASSIGNED', 'REOPENED'],
```
helfi92: Remove `keywords` and `statuses` and use `...bugzillaSearchOptions,` instead.

src/views/Project/index.jsx
helfi92: Also, let's create a constants file in the utils directory and have `good-first-bug`, `NEW`, `UNCONFIRMED`, `ASSIGNED`, `REOPENED` defined there.

```jsx
export const GOOD_FIRST_BUG = 'good-first-bug';
export const BUGZILLA_STATUSES = {
  NEW: 'NEW',
  ...
}
```

src/views/Project/index.jsx
```diff
@@ -22,15 +25,58 @@ const tagReposMapping = repositories =>
...
+      },
+    }) => ({
+      variables: {
+        search: {
+          keywords: ['good-first-bug'],
+          products: productsList(projects[project].products).length
+            ? productsList(projects[project].products)
+            : Object.keys(projects[project].products[0])[0],
+          components: productsList(projects[project].products).length
+            ? undefined
```
helfi92: Can we use `null` instead?

src/views/Project/index.jsx
```diff
@@ -22,15 +25,58 @@ const tagReposMapping = repositories =>
...
+            : Object.values(projects[project].products[0])[0],
```
helfi92: Why are we only picking the zero index element? What about the rest?

fiennyangeln: the rest is picked on the fetchMore.

helfi92: Why not pick everything from the start?

fiennyangeln: because the component is not unique for different products, e.g Bugzilla product has Administration component, and bugzilla.mozilla.org also has Administration component, then if it only specify Administration component in Bugzilla, when we combine the product and component, it may also return Administration component in bugzilla.mozilla.org

src/views/Project/index.jsx
```diff
@@ -57,18 +73,16 @@ const tagReposMapping = repositories =>
     }) => ({
       variables: {
         search: {
-          keywords: ['good-first-bug'],
+          ...bugzillaSearchOptions,
           products: productsList(projects[project].products).length
```
helfi92: `productsList` seems to check if products have no components listed under them and return the product names in an array. Would something like `productsWithNoComponents` be a better name than `productsList`?

Also, have we considered the case of products with and without components?

_Example:_

```yaml
name: Bugzilla
description: Get involved with the [Bugzilla](https://wiki.mozilla.org/Bugzilla) project or [BMO](https://wiki.mozilla.org/BMO), Mozilla's custom version
products:
 - Testing
 - Bugzilla: ['Administration', 'Attachments & Requests', 'Bug Import/Export & Moving', 'Bugzilla-General', 'bugzilla.org', 'Creating/Changing Bugs', 'Documentation', 'Email Notifications', 'Extensions', 'QA Test Scripts', 'Query/Bug Lists', 'Testing Suite', 'User Accounts', 'User Interface', 'WebService', 'Whining']
 - bugzilla.mozilla.org: ['Administration', 'API', 'Bugzilla Tweaks', 'Developer Box', 'Extensions: AntiSpam', 'Extensions: BMO', 'Extensions: ComponentWatching', 'Extensions: MyDashboard', 'Extensions: Needinfo', 'Extensions: UserProfile', 'General', 'rbbz', 'Sandstone/Mozilla Skin', 'User Interface']
```

_Note: I added a `Testing` product without components._

Refreshing the page doesn't seem to always show me the full list of bugs. Sometimes I only get bugs for Testing and sometimes I get bugs for all Testing, Bugzilla, and bugzilla.mozilla.org.

src/views/Project/index.jsx
```diff
@@ -22,15 +41,57 @@ const tagReposMapping = repositories =>
...
+        search: {
+          ...bugzillaSearchOptions,
+          products: productsWithNoComponents(projects[project].products).length
+            ? productsWithNoComponents(projects[project].products)
+            : Object.keys(projects[project].products[0])[0],
+          components: productsWithNoComponents(projects[project].products)
+            .length
+            ? undefined
+            : Object.values(projects[project].products[0])[0],
+        },
```
helfi92: A neater way than doing `components: null` would be not to provide the `components` at all:

```jsx
search: {
  ...bugzillaSearchOptions,
  ...(productsWithNoComponents(projects[project].products).length
    ? { products: productsWithNoComponents(projects[project].products) }
    : {
        products: Object.keys(projects[project].products[0])[0],
        components: Object.values(projects[project].products[0])[0],
      }),
},
```

fiennyangeln:
@helfi92 i can reproduce the error, stranger thing is that when i add `- Testing` to both bugzilla and a11y, sometimes a11y also return bugzilla result.

src/views/Project/index.jsx
```diff
@@ -22,15 +41,57 @@ const tagReposMapping = repositories =>
...
+@compose(
+  graphql(issuesQuery, {
+    skip: ({
+      match: {
+        params: { project },
+      },
+    }) => !projects[project].repositories,
+    name: 'github',
+    options: () => ({
+      fetchPolicy: 'network-only',
```
helfi92: This probably needs to be `no-cache` right?

fiennyangeln: @helfi92 but this one is for the github?

helfi92: Oh right. I guess we can leave this one `network-only`.

fiennyangeln:
@helfi92 i have finished rebasing it, but i had one issue, if there is only products with no component (meaning it wont run the fetchMore) the data wont get passed as props.  I realized it when i printed the number of bug returned in the bugzilla-graphql-gateway and even so, for Taskcluster, it keeps fetching 0 bug while in bugzilla-graphql-gateway it has several bugs.

djmitche:
How long is the data cached?  Would a ctrl-r clear the cache?  The data for this site doesn't change terribly quickly, so caching shouldn't be a big issue.

fiennyangeln:
I dont think it expire https://github.com/apollographql/apollo-cache-persist/issues/53 but i will try to find other resources. @djmitche

fiennyangeln:
@helfi92 You're right! the network-only can work if we force it to wait for the first result before fetching more. I added a method to check for it. Wdyt?

src/views/Project/index.jsx
```diff
@@ -71,19 +132,72 @@ export default class Project extends Component {
...
-  fetch = searchQuery => {
+  loadMore = () => {
+    if (this.props.bugzilla && this.props.bugzilla.loading) {
+      setTimeout(this.loadMore, 50);
+
+      return;
+    }
+
+    this.load();
+  };
```
helfi92: You can use a React lifecycle method to detect prop changes instead of using `setTimeout`. Remove `loadMore` and `componentDidMount` and give the following a try:

```jsx
componentDidUpdate(prevProps) {
  if (prevProps.bugzilla.loading && !this.props.bugzilla.loading) {
    this.load();
  }
}
```

fiennyangeln: Ok! Added some logic to handle case where no bugzilla product exist (all github query)

helfi92:
This needs a rebase.

fiennyangeln:
@helfi92 ok done!

src/views/Project/index.jsx
```diff
@@ -32,15 +51,57 @@ const tagReposMapping = repositories =>
...
+      },
+    }) => ({
+      fetchPolicy: 'network-only',
+      variables: {
+        search: {
+          ...bugzillaSearchOptions,
+          ...(productsWithNoComponents(projects[project].products).length
+            ? { products: productsWithNoComponents(projects[project].products) }
+            : {
+                products: Object.keys(projects[project].products[0])[0],
```
helfi92: Could you add a comment explaining **why** we are only grabbing the first entry of the array? This will help people hacking on this in the future to quickly understand.

src/views/Project/index.jsx
```diff
@@ -32,15 +51,57 @@ const tagReposMapping = repositories =>
...
+                components: Object.values(projects[project].products[0])[0],
```
helfi92: Could you add a comment explaining **why** we are only grabbing the first entry of the array? This will help people hacking on this in the future to quickly understand.

src/views/Project/index.jsx
```diff
@@ -113,22 +225,34 @@ export default class Project extends Component {
...
+      project.products
+        ? project.products.filter(product => typeof product !== 'string')
+        : []
+    );
+
+    await Promise.all(
+      Object.entries(productWithComponentList).map(([products, components]) =>
+        this.fetchBugzilla([products], components)
+      )
+    );
```
helfi92: Could you add a comment explaining **why** we are only fetching bugs from products with components but not products without components? This will help people hacking on this in the future to quickly understand.

fiennyangeln:
@helfi92 i added the comment!

helfi92:
Opened https://github.com/mozilla-frontend-infra/bugzilla-graphql-gateway/issues/3 to see how we are going to access the bugzilla graphql endpoint in production.

README.md
```diff
@@ -55,7 +56,13 @@ Install npm dependencies and start it up:
 - `yarn`
 - `yarn start`
 
-This will start a local development server on port 5000 (http://localhost:5000).
+This will start a local development server on port 5000 (http://localhost:5000). 
+
+**Note**: For fetching from Bugzilla, this application rely on
```
helfi92: Add a new section `Web Server`before `Environment`. You can have something like:

Codetribute relies on two servers, namely, GitHub’s GraphQL API v4 (managed by GitHub) and [bugzilla-graphql-gateway](http://github.com/mozilla-frontend-infra/bugzilla-graphql-gateway). The latter is required to perform queries to the Bugzilla API. For a local setup, clone the repo and follow the instruction for starting it prior to launching this application. You will need to launch the bugzilla-graphql-gateway in a terminal instance separate from this application in order to run both simultaneously.

PR #55, opened 2018-06-25
Fix sorting for tag by providing support for array type

### LINKED ISSUE #54: Sorting by tag not working
Clicking on the "Tag" column throws an error in the console: `referenceElement.match is not a function`.

fiennyangeln:
Fix #54

### DIFF
--- a/src/components/DataTable/index.jsx
+++ b/src/components/DataTable/index.jsx
@@ -54,7 +54,7 @@ class DataTable extends Component {
   };
 
   handleHeaderClick = ({ target }) => {
-    if (this.props.onHeaderClick && target.id !== 'Tags') {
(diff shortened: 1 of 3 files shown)


src/components/DataTable/index.jsx
```diff
@@ -54,7 +54,7 @@ class DataTable extends Component {
   };
 
   handleHeaderClick = ({ target }) => {
-    if (this.props.onHeaderClick) {
+    if (this.props.onHeaderClick && target.id !== 'Tags') {
```
helfi92: `DataTable` should not have knowldege of `Tags`. Put this logic in `TasksTable`.

helfi92:
Merge when ready.

PR #56, opened 2018-06-25
Display project introduction in a expansion panel

### LINKED ISSUE #53: Have the project introduction be in a panel
It would be neat to be able to hide the project introduction if a user does not want to be bothered by it. Material UI offers the [expansion panel](https://material-ui.com/demos/expansion-panels/) which we could make use of it. When a user enters the page, we could have the panel expanded by default.

fiennyangeln:
Uses expansion panel to display project info
Fix #53

### DIFF
--- a/src/views/Project/index.jsx
+++ b/src/views/Project/index.jsx
@@ -121,9 +121,9 @@ export default class Project extends Component {
             {project.name}
           </Typography>
           {project.introduction && (
-            <ExpansionPanel>
+            <ExpansionPanel defaultExpanded>
               <ExpansionPanelSummary expandIcon={<ChevronDownIcon />}>
-                <Typography>Project Introduction</Typography>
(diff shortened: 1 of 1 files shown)


fiennyangeln:
@helfi92 how about displaying it in a full screen dialog? https://material-ui.com/demos/dialogs/

src/views/Project/index.jsx
```diff
@@ -116,14 +120,21 @@ export default class Project extends Component {
           <Typography variant="display2" align="center">
             {project.name}
           </Typography>
+          {project.introduction && (
+            <ExpansionPanel>
+              <ExpansionPanelSummary expandIcon={<ChevronDownIcon />}>
+                <Typography>Project Introduction</Typography>
```
helfi92: I think we want a `variant="display1"` to make sure `Project Introduction` is bigger in size than markdown.

PR #57, opened 2018-06-26
Display unassigned bugs by default

### LINKED ISSUE #38: Display unassigned bugs by default

fiennyangeln:
Add switch to which is defaulted to false (not display assigned bugs)

Fix #38


fiennyangeln:
@helfi92 I initially want to use it but i think it might be hard to find since the table has longer length than the screen sometimes. can we put it on the left?

helfi92:
If you look at the example from the link I provided, you will see the table has a longer length (requires the user to scroll horizontally to see all of the columns) but the filter icon is always visible since it sits on top of the Table component.

fiennyangeln:
@helfi92 oh okok thanks!

src/components/DataTable/index.jsx
```diff
@@ -1,19 +1,29 @@
...
 import TableBody from '@material-ui/core/TableBody';
 import TableCell from '@material-ui/core/TableCell';
 import TableHead from '@material-ui/core/TableHead';
 import TableRow from '@material-ui/core/TableRow';
 import TableSortLabel from '@material-ui/core/TableSortLabel';
-import { arrayOf, func, string, oneOf, object } from 'prop-types';
+import Typography from '@material-ui/core/Typography';
+import IconButton from '@material-ui/core/IconButton';
+import FilterVariantIcon from 'mdi-react/FilterVariantIcon';
+import Toolbar from '@material-ui/core/Toolbar';
```
helfi92: Move `Toolbar` on top of `FilterVariantIcon` so that `material-ui` packages are grouped together.

src/components/DataTable/index.jsx
```diff
@@ -67,40 +92,56 @@ class DataTable extends Component {
...
-              ))}
-            </TableRow>
-          </TableHead>
+      <Fragment>
+        {title && (
+          <Toolbar>
+            <Typography variant="title" className={classes.title}>
+              {title}
+            </Typography>
+            <div className={classes.spacer} />
```
helfi92: Let's not use empty `div`s.

src/components/DataTable/index.jsx
```diff
@@ -59,6 +78,12 @@ class DataTable extends Component {
     }
   };
 
+  handleFilterClick = () => {
+    if (this.props.onFilterClick) {
+      this.props.onFilterClick();
```
helfi92: Clicking on the filter icon should open a menu item with "Unassigned" as a menu item. Clicking on "Unassigned" should show bugs/issues that are unassigned.

fiennyangeln: Right now it opens the filter form which has a switch with the word "Also display assigned bugs"  since the default is to display the unassigned bugs , do you mean using drop down button with Unassigned as the default selected?

helfi92: Not necessarily. It should be okay having "Unassigned" not selected by default but let the user select it if they can't find unassigned bugs/issues.

src/components/DataTable/index.jsx
```diff
@@ -46,11 +53,20 @@ class DataTable extends Component {
...
+    /**
+     * A function to execute when the filter icon is clicked.
+     */
+    onFilterClick: func,
   };
 
   static defaultProps = {
     sortByHeader: null,
     sortDirection: 'desc',
+    filters: undefined,
```
helfi92: `filters: null,`

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+export default class FilterMenu extends Component {
+  state = {
+    anchorEl: null,
+  };
+
+  getSelectedAssignee() {
+    const { assignee } = this.getQuery();
+
+    return assigneeList.indexOf(assignee) > -1 ? assignee : ASSIGNEE.UNASSIGNED;
+  }
```
helfi92: A component should not be getting data from the URL rather it should be getting them from props. We should define props for this component.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+  getQuery() {
+    const { location } = this.props;
+    const query = parse(location.search.slice(1));
+
+    return query;
+  }
```
helfi92: Remove `getQuery` and use `props` instead.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+  handleClickListItem = event => {
```
helfi92: Use the format `handle${subject}Click`. In our case maybe something like `handleAssigneeClick`.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+  handleMenuItemClick = (event, assignee) => {
+    this.setState({ anchorEl: null });
+    this.props.history.push({
+      search: `?${stringify({
+        ...this.getQuery(),
+        assignee,
+      })}`,
+    });
+  };
```
helfi92: If we are going to make this file a component then we shouldn't be making URL changes in the component. It should be handled by the parent view.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+        <Menu
+          id="lock-menu"
+          anchorEl={anchorEl}
+          open={Boolean(anchorEl)}
+          onClose={this.handleClose}>
+          {assigneeList.map(assignee => (
+            <MenuItem
+              key={assignee}
+              selected={assignee === selectedAssignee}
+              onClick={event => this.handleMenuItemClick(event, assignee)}>
```
helfi92: I'm not sure having two `onClick` (one on `ListItem` and one here) is necessary. Have a look at `TextField` [menu example](https://material-ui.com/demos/text-fields/#textfield) where they show how you can do everything using one onChange handler.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+const assigneeList = Object.values(ASSIGNEE);
+
+@withRouter
+@withStyles(theme => ({
+  root: {
+    margin: `0px ${3 * theme.spacing.unit}px`,
+    maxWidth: 360,
+  },
+}))
+export default class FilterMenu extends Component {
```
helfi92: The directory name is `FilterForm` but the component name is `FilterMenu`. They should be the same.

src/components/TasksTable/index.jsx
```diff
@@ -36,6 +38,10 @@ const sorted = pipe(
   },
 }))
 export default class TasksTable extends Component {
+  state = {
+    displayFilter: false,
```
helfi92: Rename to `showFilterContent`.

src/utils/constants.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable */
+
+export const ASSIGNEE = {
+  ANY: 'Any',
+  UNASSIGNED: 'Unassigned',
+  ASSIGNED: 'Assigned',
+};
```
helfi92: 👍

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+    return assigneeList.indexOf(assignee) > -1 ? assignee : ASSIGNEE.UNASSIGNED;
+  }
+
+  handleFilterItemClick = event => {
+    this.props.onFilterItemClick({ assignee: event.target.value });
+  };
+
+  render() {
+    const selectedAssignee = this.getSelectedAssignee();
```
helfi92: No need to have a method for that. Use it inside the `render` method.

```jsx
const assignment = assigneeList.includes(assignee) ? assignee : ASSIGNEE.UNASSIGNED;
```

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+    const { classes } = this.props;
+
+    return (
+      <div className={classes.root}>
+        <TextField
+          select
+          label="Assignee"
+          value={selectedAssignee}
+          onChange={this.handleFilterItemClick}>
```
helfi92: Rename to `handleFilterChange`.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,60 @@
...
+          {assigneeList.map(assignee => (
+            <MenuItem key={assignee} value={assignee}>
+              {assignee}
+            </MenuItem>
+          ))}
+        </TextField>
+      </div>
+    );
+  }
+}
```
helfi92: This component simply renders a `TextField`. We don't need a filter component for that. We can just use `TextField` directly inside `TasksTable`.

src/components/TasksTable/index.jsx
```diff
@@ -33,6 +35,10 @@ const sorted = pipe(
   },
 }))
 export default class TasksTable extends Component {
+  state = {
+    showFilterContent: false,
```
helfi92: Will assignee be the only filter option we show? If this is the case, then maybe when a user clicks on the filter icon, we display the `<TextField select ... />` directly. If there's only one filter planned (at least for the MVP) then we don't need to have the dedicated filter content area.

To show which filter is selected, I can see us putting a check (✓) icon beside the menu item name.

helfi92: In that case, maybe having the dedicated filter content area makes sense.

src/components/FilterForm/index.jsx
fiennyangeln: if there is two `TextField`, wdyt? Or should i save it for later?

helfi92: For now let's remove this component and just inline it in `TasksTable`.

src/components/TasksTable/index.jsx
```diff
@@ -70,11 +86,28 @@ export default class TasksTable extends Component {
...
     };
   }
 
+  setQuery = query => {
+    this.props.history.push({
+      search: `?${stringify(query)}`,
+    });
+  };
+
+  handleFilterClick = () => {
```
helfi92: Let's rename this function to `handleFilterToggle`.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,66 @@
+import { Component } from 'react';
+import { withRouter } from 'react-router-dom';
+import MenuItem from '@material-ui/core/MenuItem';
+import TextField from '@material-ui/core/TextField';
+import { withStyles } from '@material-ui/core/styles';
+import CheckIcon from 'mdi-react/CheckIcon';
```
helfi92: Now that we will keep the filter area content, we won't need the check icon. Let's bring it back to how it was previously.

src/components/FilterForm/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+import { func, string } from 'prop-types';
+import { ASSIGNEE } from '../../utils/constants';
+
+const assigneeList = Object.values(ASSIGNEE);
```
helfi92: Rename to `assignments`.

src/components/DataTable/index.jsx
```diff
@@ -74,16 +94,23 @@ class DataTable extends Component {
...
       items,
       title,
+      filters,
     } = this.props;
     const colSpan = (headers && headers.length) || 0;
 
     return (
       <Fragment>
         {title && (
           <Toolbar>
```
helfi92:
```jsx
toolbar: {
  justifyContent: 'space-between',
},


// ...
```

src/components/DataTable/index.jsx
```diff
@@ -74,16 +94,23 @@ class DataTable extends Component {
...
-            <Typography variant="title">{title}</Typography>
+            <Typography variant="title" className={classes.title}>
```
helfi92: Remove `className={classes.title}`.

src/components/TasksTable/index.jsx
```diff
@@ -31,8 +35,18 @@ const sorted = pipe(
   tags: {
     whiteSpace: 'nowrap',
   },
+  dropdown: {
+    minWidth: 200,
+  },
+  filter: {
+    padding: `0px ${3 * theme.spacing.unit}px`,
```
helfi92: On smaller screens this will not align with "Bugs & Issues" because the padding changes with screen size. To fix this you can do:

```jsx
filter: {
  ...theme.mixins.gutters(),
},
```

src/components/DataTable/index.jsx
```diff
@@ -74,16 +94,21 @@ class DataTable extends Component {
...
     const colSpan = (headers && headers.length) || 0;
 
     return (
       <Fragment>
         {title && (
-          <Toolbar>
+          <Toolbar className={classes.toolbar}>
             <Typography variant="title">{title}</Typography>
+            <IconButton onClick={this.handleFilterClick}>
+              <FilterVariantIcon onClick={this.handleFilterClick} />
```
helfi92: Remove the `onClick` from `FilterVariantIcon` since we already have it on `IconButton`.

PR #58, opened 2018-06-27
Add filter based on the tags

### LINKED ISSUE #19: Add filtering for the bug table based on tag
When user clicks on the chip, it should filter the bugs that has that tag

fiennyangeln:
Make the chips clickable and when click, filter to only display bug with that tag, when clicked again, it will display all bugs

Fix #19

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -40,9 +40,9 @@ const assignments = Object.values(ASSIGNEE);
     overflowX: 'auto',
   },
   clickedChip: {
-    backgroundColor: theme.palette.primary.main,
+    backgroundColor: theme.palette.secondary.dark,
     '&:focus': {
-      backgroundColor: theme.palette.primary.main,
+      backgroundColor: theme.palette.secondary.dark,
     },
   },
   tags: {
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -9,8 +9,9 @@ import { camelCase } from 'change-case';
 import { formatDistance } from 'date-fns';
 import MenuItem from '@material-ui/core/MenuItem';
 import TextField from '@material-ui/core/TextField';
-import { memoizeWith, pipe, sort as rSort, map } from 'ramda';
+import { memoizeWith, omit, pipe, sort as rSort, map } from 'ramda';
 import { stringify, parse } from 'qs';
+import classNames from 'classnames';
 import DataTable from '../DataTable';
 import sort from '../../utils/sort';
 import { ASSIGNEE } from '../../utils/constants';
@@ -31,6 +32,12 @@ const assignments = Object.values(ASSIGNEE);
   },
   link: {
     textDecoration: 'none',
+  },
+  chip: {
+    marginRight: 1,
+  },
+  clickedChip: {
+    backgroundColor: theme.palette.primary.main,
   },
   tags: {
     whiteSpace: 'nowrap',
@@ -56,12 +63,12 @@ export default class TasksTable extends Component {
   };
 
   getTableData = memoizeWith(
-    (sortBy = 'Last Updated', sortDirection = 'desc', items, assignee) => {
+    (sortBy, sortDirection, tag, items, assignee) => {
       const ids = sorted(items);
 
-      return `${ids.join('-')}-${sortBy}-${sortDirection}-${assignee}`;
+      return `${ids.join('-')}-${sortBy}-${sortDirection}-${tag}-${assignee}`;
     },
-    (sortBy = 'Last Updated', sortDirection = 'desc', items, assignee) => {
+    (sortBy, sortDirection, tag, items, assignee) => {
       const sortByProperty = camelCase(sortBy);
       let filteredItems = [];
 
@@ -73,7 +80,9 @@ export default class TasksTable extends Component {
         filteredItems = items.filter(item => item.assignee === '-');
       }
 
-      return [...filteredItems].sort((a, b) => {
+      return [
+        ...filteredItems.filter(item => !tag || item.tags.indexOf(tag) > -1),
+      ].sort((a, b) => {
         const firstElement =
           sortDirection === 'desc' ? b[sortByProperty] : a[sortByProperty];
         const secondElement =
@@ -119,14 +128,24 @@ export default class TasksTable extends Component {
     this.setQuery({ ...query, sortBy, sortDirection });
   };
 
+  handleChipClick = tag => event => {
+    const query = this.getQuery();
+    const newQuery =
+      query.tag === tag ? omit(['tag'], query) : { ...query, tag };
+
+    this.setQuery(newQuery);
+
+    event.preventDefault();
+  };
+
   render() {
     const { items, classes } = this.props;
     const { showFilterContent } = this.state;
-    const { sortBy, sortDirection, assignee } = this.getQuery();
+    const { sortBy, sortDirection, tag, assignee } = this.getQuery();
     const assignment = assignments.includes(assignee)
       ? assignee
       : ASSIGNEE.UNASSIGNED;
-    const data = this.getTableData(sortBy, sortDirection, items, assignee);
+    const data = this.getTableData(sortBy, sortDirection, tag, items, assignee);
 
     return (
       <div className={classes.tableWrapper}>
@@ -149,7 +168,14 @@ export default class TasksTable extends Component {
               <TableCell className={classes.summary}>{item.summary}</TableCell>
               <TableCell className={classes.tags}>
                 {item.tags.map(tag => (
-                  <Chip key={tag} label={tag} className={classes.chip} />
+                  <Chip
+                    key={tag}
+                    label={tag}
+                    className={classNames(classes.chip, {
+                      [classes.clickedChip]: tag === query.tag,
+                    })}
+                    onClick={this.handleChipClick(tag)}
+                  />
                 ))}
               </TableCell>
               <TableCell>{item.assignee}</TableCell>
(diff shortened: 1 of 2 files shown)


helfi92:
A rebase is needed.

src/components/TasksTable/index.jsx
```diff
@@ -28,6 +29,12 @@ const sorted = pipe(
   link: {
     textDecoration: 'none',
   },
+  chip: {
+    marginRight: 1,
+  },
```
helfi92: This style is defined in `theme.jsx`. We should remove it from here.

src/components/TasksTable/index.jsx
```diff
@@ -42,19 +49,21 @@ export default class TasksTable extends Component {
...
       const sortByProperty = camelCase(sortBy);
 
       if (!sortBy) {
         return items;
       }
 
-      return [...items].sort((a, b) => {
+      return [
+        ...items.filter(item => !tag || item.tags.indexOf(tag) > -1),
+      ].sort((a, b) => {
```
helfi92: The filter() method creates a new array so we don't need the `[...items]` trick anymore. Also, use `.includes()` instead of `indexOf()` when you can.

```jsx
items
  .filter(item => !tag || item.tags.includes(tag))
  .sort(...)
```

src/components/TasksTable/index.jsx
```diff
@@ -92,10 +102,23 @@ export default class TasksTable extends Component {
     });
   };
 
+  handleChipClick = tag => event => {
+    const query = this.getQuery();
+    const newQuery =
+      query.tag === tag ? omit(['tag'], query) : { ...query, tag };
```
helfi92: Why are we omitting the `tag` key? Also, this should probably be logic handled by `getQuery`.

fiennyangeln:
@helfi92 I had to change the ErrorPanel padding as it gives me error when I want to commit ( ask me to give `( 2 * theme.spacing.unit)`, then after I give it, it gives lint error in the netlify deploy.. Should I make separate issue for it? 
In the theme.jsx I had to add the style for clickable focused to be similar as the default one so that when user click 'good-first-bug' in bug1 then 'good-first-bug' in bug2, the one in bug2 can have same color as the unclicked one (since it means good-first-bug has been deselected)

src/components/TasksTable/index.jsx
```diff
@@ -92,33 +104,54 @@ export default class TasksTable extends Component {
...
               </TableCell>
-              <TableCell className={classes.summary}>{item.summary}</TableCell>
+              <TableCell
+                className={classes.summary}
+                target="_blank"
+                rel="noopener noreferrer"
+                component="a"
+                href={item.url}>
+                {item.summary}
+                <LinkIcon size={iconSize} className={classes.icon} />
```
helfi92: Did you look at tc-web on how we handle table cells which have links?

_Screenshots from the Worker Types view (http://localhost:5080/provisioners/aws-provisioner-v1)_

screen shot 2018-06-29 at 8 59 26 am

screen shot 2018-06-29 at 8 59 56 am

src/components/TasksTable/index.jsx
```diff
@@ -92,33 +104,54 @@ export default class TasksTable extends Component {
...
                 {item.tags.map(tag => (
-                  <Chip key={tag} label={tag} className={classes.chip} />
+                  <Chip
+                    key={tag}
+                    label={tag}
+                    className={classNames({
+                      [classes.clickedChip]: tag === query.tag,
+                    })}
+                    onClick={this.handleChipClick(tag)}
+                  />
```
helfi92: Hovering over the chip makes the chip grey with white text. Going from green-blue to grey on hover is not a smooth transition.In addition, gray on white text is not enough contrast making it hard for the eye to read the text. Could we instead just use a different shade of the green-blue color? You can try exporting the `emphasize` function from the library and provide it with the current chip color.

To export the `emphasize` function, you will do `import { emphasize } from '@material-ui/core/styles/colorManipulator';`

https://github.com/mui-org/material-ui/blob/8bc7d11c9a445de5acbf0de8ec550e3bdd23c565/packages/material-ui/src/Chip/Chip.js#L42-L44

src/components/ErrorPanel/index.jsx
```diff
@@ -12,7 +12,7 @@ import ErrorBox from './ErrorBox';
 
 @withStyles(theme => ({
   paper: {
-    padding: `0 ${2 * theme.spacing.unit / 3}px`,
+    padding: `0 ${theme.spacing.unit}px`,
```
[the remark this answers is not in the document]
fiennyangeln: it is something like this (in the deploy preview in PR #64 https://app.netlify.com/sites/codetribute/deploys/5b35ffc667610c07eb11dcb1 ) but if i try to commit it without adding ( ), the commit will fail.
screen shot 2018-06-30 at 12 56 48 am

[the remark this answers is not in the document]
helfi92: The error is recommending you wrap the multiplication with parenthesis. Did you try

``` `0 ${(2 * theme.spacing.unit) / 3}px`, ```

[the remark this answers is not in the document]
Comment (fiennyangeln) [reply src/components/ErrorPanel/index.jsx]: @helfi92 If I do it, then the deploy preview in netlify will fail, is it actually ok?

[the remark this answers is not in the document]
Comment (helfi92) [reply src/components/ErrorPanel/index.jsx]: If you want to know if netlify will fail you can try running `yarn build` from your terminal and see if it passes. When that passes then netlify will most probably pass. Also, `2/3` seems weird for material padding. Can we try using `padding: ${2 * theme.spacing.unit}` just like tc-web.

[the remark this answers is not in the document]
Comment (fiennyangeln) [reply src/components/ErrorPanel/index.jsx]: ok sure, but the spacing unit isn't overrided in here unlike in tc-web

File: src/components/TasksTable/index.jsx [hunk 7]
```diff
@@ -18,19 +24,39 @@ const sorted = pipe(
...
+  },
+  summaryButton: {
+    '& svg': {
+      transition: transitions.create('fill'),
+      fill: theme.palette.primary.main,
+    },
+    '&:hover svg': {
+      fill: 'white',
+    },
+  },
```
helfi92: This is good to have when using something like a dark theme. Having a gray background with white text makes it hard to see so let's keep the default behavior. Remove `summaryButton`.

src/components/TasksTable/index.jsx
```diff
@@ -92,33 +121,62 @@ export default class TasksTable extends Component {
...
-            <TableRow
-              hover
-              target="_blank"
-              rel="noopener noreferrer"
-              tabIndex={-1}
-              key={item.summary}
-              component="a"
-              href={item.url}
-              className={classes.link}>
+            <TableRow hover tabIndex={-1} key={item.summary}>
```
helfi92: Remove the `hover` property.

src/theme.jsx
```diff
@@ -53,9 +54,17 @@ const theme = createMuiTheme({
...
         color: 'white',
       },
+      clickable: {
+        '&:focus': {
+          backgroundColor: PRIMARY.MAIN,
+        },
+        '&:hover': {
+          backgroundColor: emphasize(SECONDARY.DARK),
+        },
+      },
```
helfi92: Since we're using a LinkIcon with hover on the summary column, we can remove the hover that is applied on the full row. This will also help with the issue of being navigated away from the site when clicking outside the chip area by accident.

Remove `MuiTableRow`.

src/components/TasksTable/index.jsx
```diff
@@ -2,13 +2,19 @@ import { Component } from 'react';
...
+import LinkIcon from 'mdi-react/LinkIcon';
 import { withRouter } from 'react-router-dom';
 import { arrayOf, object } from 'prop-types';
 import { camelCase } from 'change-case';
 import { formatDistance } from 'date-fns';
-import { memoizeWith, pipe, sort as rSort, map } from 'ramda';
+import { memoizeWith, omit, pipe, sort as rSort, map } from 'ramda';
 import { stringify, parse } from 'qs';
+import classNames from 'classnames';
+import transitions from '@material-ui/core/styles/transitions';
```
helfi92: Remove the `transitions` import. See reason in other comment.

src/theme.jsx
```diff
@@ -53,9 +54,17 @@ const theme = createMuiTheme({
       root: {
         height: 24,
         marginRight: 1,
```
helfi92: Let's change this to `4` to make sure we don't accidentally click on another chip.

src/components/TasksTable/index.jsx
```diff
@@ -92,33 +111,59 @@ export default class TasksTable extends Component {
...
               <TableCell className={classes.tags}>
                 {item.tags.map(tag => (
-                  <Chip key={tag} label={tag} className={classes.chip} />
+                  <Chip
+                    key={tag}
+                    label={tag}
+                    className={classNames({
+                      [classes.clickedChip]: tag === query.tag,
+                    })}
+                    onClick={this.handleChipClick(tag)}
```
helfi92: For every tag, this is invoking `handleChipClick(tag)` on page load. Another way to do this is by adding a name attribute `<Chip name={tag} onClick={this.handleTagClick} ...>` and changing `handleChipClick` to be:

```jsx
handleTagClick = ({ currentTarget }) => {
  const tag = currentTarget.getAttribute('name');
  const query = this.getQuery();
  const newQuery =
    query.tag === tag ? omit(['tag'], query) : { ...query, tag };

  this.props.history.push({
    search: `?${stringify(newQuery)}`,
  });
};
```

Doing it like that will only trigger that function when a tag is actually clicked.

helfi92:
Feel free to merge :)

Issue #60, opened 2018-06-27
Add another way to display bug/issue in the form of Card


fiennyangeln:

helfi92:
https://github.com/mozilla-frontend-infra/codetribute/pull/64#issuecomment-405930800

PR #64, opened 2018-06-29
add options to display card

fiennyangeln:
- I make the `TasksList` component to be called by TasksTable, not views/project as I realized some of the method are similar (getTableData), the filter button, the view options button, and the filters later
- I make the options to display available only for large screen, for small screen, it will only display card

Bug #60


src/components/TasksTable/index.jsx
```diff
@@ -92,49 +107,86 @@ export default class TasksTable extends Component {
...
+      <Fragment>
+        <IconButton value onClick={this.handleViewOptionsClick}>
+          <TableColumnIcon />
+        </IconButton>
+        <IconButton value={false} onClick={this.handleViewOptionsClick}>
+          <TableLargeIcon />
+        </IconButton>
+      </Fragment>
+    );
+  }
```
helfi92: Instead of having two icons, we can have 1 icon only. When the bugs/issues are displayed in a table format, we show the card icon. When the bugs/issues are displayed in a card format, we show the table icon.

src/components/TasksTable/index.jsx
```diff
@@ -143,9 +157,25 @@ export default class TasksTable extends Component {
...
+
+  renderViewOptions() {
+    const { displayCard } = this.state;
+
+    return (
+      <Fragment>
+        <IconButton value={!displayCard} onClick={this.handleViewOptionsClick}>
+          {displayCard ? <TableLargeIcon /> : <TableColumnIcon />}
+        </IconButton>
+      </Fragment>
```
helfi92: Remove `Fragment`.

src/components/DataTable/index.jsx
```diff
@@ -102,6 +103,7 @@ class DataTable extends Component {
         {title && (
           <Toolbar className={classes.toolbar}>
```
helfi92: We should probably take out the `Toolbar` component outside and let `DataTable` only take care of rendering a table. Are we able to do this?

```jsx
{
 displayTableLayout ?
   <DataTable ...> :
   <DataCard ...>
}
```

helfi92: We should also have the table icon be beside the filter icon instead of being by itself in the center:

screen shot 2018-07-05 at 9 50 58 am

src/components/TasksTable/index.jsx
```diff
@@ -225,105 +286,150 @@ export default class TasksTable extends Component {
...
+                  component="a"
+                  href={item.url}
+                  className={classes.link}>
+                  {item.summary}
+                  <LinkIcon size={14} />
+                </Typography>
+                <Divider light />
+                <Typography component="p" gutterBottom>
+                  Project: <strong>{item.project}</strong> | Assignee:{' '}
+                  <strong>{item.assignee}</strong> | Last Update:{' '}
```
helfi92: `s/Update/Updated/`

fiennyangeln:
@helfi92 how about putting the Feeling Adventurous button as a footer with more words about what it is? but im not sure how it will look like on mobile

helfi92:
I don't think we need more words about what that button does. If the user is adventurous, they will click on the button and find out what it does :)

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+  },
+  '@media screen and (min-width: 900px)': {
+    paper: {
+      width: 900,
+    },
+  },
+  title: {
+    flex: '1 1 100%',
+  },
```
helfi92: This style is not being used. Remove it.

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { withStyles } from '@material-ui/core/styles';
+import Grid from '@material-ui/core/Grid';
+import { arrayOf, object } from 'prop-types';
+
+@withStyles(theme => ({
+  paper: {
+    minHeight: 100,
+    paddingBottom: theme.spacing.unit,
+  },
+  '@media screen and (max-width: 900px)': {
```
helfi92: Use material UI breakpoints instead of hardcoding values here: https://material-ui.com/layout/breakpoints/#breakpoints.

If you are unsatisfied with the default breakpoints, you can always configure them in `theme.jsx`.

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+    paper: {
+      width: '90vw',
+    },
+  },
+  '@media screen and (min-width: 900px)': {
```
helfi92: Use material breakpoints instead of hardcoding values here.

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+  },
+  title: {
+    flex: '1 1 100%',
+  },
+}))
+export default class TasksList extends Component {
+  static propTypes = {
+    /**
+     * A list of objects to display. Each element in the list is represented
```
helfi92: s/objects/tasks/

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  render() {
+    const { data, classes, renderRow } = this.props;
+
+    return (
+      <Grid container direction="column" alignItems="center" spacing={16}>
+        {data.map(item => (
+          <Grid item xs={12} key={item.summary}>
+            <Card raised className={classes.paper}>
```
helfi92: Try removing the `raised` prop. The shadows its creating are not sitting gracefully in the app imho.

src/components/TasksTable/index.jsx
```diff
@@ -327,6 +383,28 @@ export default class TasksTable extends Component {
...
+                      <ArrowUpIcon size={iconSize} />
+                    ) : (
+                      <ArrowDownIcon size={iconSize} />
+                    ))}
+                  {sortBy &&
+                    header !== sortBy && (
+                      <Fragment>&nbsp;&nbsp;&nbsp;&nbsp;</Fragment>
+                    )}
+                  {header}
+                </MenuItem>
```
helfi92: screen shot 2018-07-16 at 9 19 55 am

* The arrow should be a little bit taller
* There should be a little bit more space between the text and the arrow icon

src/components/TasksTable/index.jsx
```diff
@@ -335,47 +413,57 @@ export default class TasksTable extends Component {
             </Button>
           </div>
         )}
-        <div className={classes.tableWrapper}>
-          <DataTable
-            items={data}
+        {displayCardLayout || this.props.width === 'xs' ? (
```
helfi92: You can probably use the `Hidden` component instead of `this.props.width === 'xs'`. https://material-ui.com/layout/hidden/#hidden

src/components/TasksTable/index.jsx
```diff
@@ -335,47 +413,57 @@ export default class TasksTable extends Component {
...
+                <Divider light />
+                <Typography
+                  component="div"
+                  gutterBottom
+                  className={classNames(
+                    classes.cardItem,
+                    classes.cardDescription
+                  )}>
+                  <div>
+                    Assignee: <strong>{item.assignee}</strong>
```
helfi92: Instead of using "Assignee", you can pick an icon fro `mdi-react`:  https://materialdesignicons.com/

For example you can do `import Account from 'mdi-react/AccountIcon'` and use that instead.

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    },
+  },
+}))
+export default class TasksList extends Component {
+  static propTypes = {
+    /**
+     * A list of tasks to display. Each element in the list is represented
+     * by a row and each element's key-value pair represents a column.
+     */
+    data: arrayOf(object),
```
helfi92: Rename to `items` to stay in sync with `DataTable`. This property should also be set as required.

src/components/DataCard/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    },
+  },
+  [theme.breakpoints.up('md')]: {
+    paper: {
+      width: 900,
+    },
+  },
+}))
+export default class TasksList extends Component {
+  static propTypes = {
```
helfi92: Add `renderRow` to `propTypes`. It should be similar to the `DataTable` component.

src/components/TasksTable/index.jsx
```diff
@@ -9,25 +9,33 @@ import ListItem from '@material-ui/core/ListItem';
...
+import AccountIcon from 'mdi-react/AccountIcon';
 import { formatDistance, differenceInCalendarDays } from 'date-fns';
-import { memoizeWith, omit, pipe, sort as rSort, map } from 'ramda';
+import ArrowDownIcon from 'mdi-react/ArrowDownIcon';
+import ArrowUpIcon from 'mdi-react/ArrowUpIcon';
 import MenuItem from '@material-ui/core/MenuItem';
 import TextField from '@material-ui/core/TextField';
+import Typography from '@material-ui/core/Typography';
+import Toolbar from '@material-ui/core/Toolbar';
+import DotsHorizontalIcon from 'mdi-react/DotsHorizontalIcon';
```
helfi92: What the "..." icon mean? The more I think about it the more I think we should always show the filtering content area. Having to click multiple buttons (a click for switching the layout and another for open the filtering content area, etc.) can be cumbersome.

I think we should remove this icon and show the filter content area by default.

src/utils/constants.js
```diff
@@ -14,3 +14,10 @@ export const ASSIGNEE = {
   ASSIGNED: 'Assigned',
 };
 export const ALL_PROJECTS = 'All';
+export const HEADERS = {
```
helfi92: Rename to `PROJECT_HEADERS`.

src/components/TasksTable/index.jsx
```diff
@@ -263,9 +304,84 @@ export default class TasksTable extends Component {
...
+  renderCard = data => {
+    const { classes } = this.props;
+    const query = this.getQuery();
+    const iconSize = 14;
+
+    return (
+      <DataCard
+        data={data}
+        renderRow={item => (
+          <div>
```
helfi92: See if we can switch this to `Fragment`. If not, it's fine.

src/components/TasksTable/index.jsx
```diff
@@ -281,22 +397,30 @@ export default class TasksTable extends Component {
...
+          <Hidden smDown>
+            <IconButton
+              value={!displayCardLayout}
+              onClick={this.handleViewOptionsClick}>
+              {displayCardLayout ? <TableLargeIcon /> : <TableColumnIcon />}
+            </IconButton>
+          </Hidden>
+          <IconButton onClick={this.handleFilterToggle}>
+            <DotsHorizontalIcon />
+          </IconButton>
```
helfi92: Remove this icon and have the filter show by default.

src/components/TasksTable/index.jsx
```diff
@@ -263,9 +304,84 @@ export default class TasksTable extends Component {
...
+                  className={classNames({
+                    [classes.clickedChip]: tag === query.tag,
+                  })}
+                  onClick={this.handleTagClick}
+                />
+              ))}
+            </div>
+          </div>
+        )}
+      />
```
helfi92: We will need a button to open the drawer as well similar to the table layout.

src/components/TasksTable/index.jsx
```diff
@@ -263,9 +304,84 @@ export default class TasksTable extends Component {
...
+                <LinkIcon size={iconSize} />
+              </ListItem>
+            </List>
+            <Divider light />
+            <Typography
+              component="div"
+              gutterBottom
+              className={classNames(classes.cardItem, classes.cardDescription)}>
+              <div>
+                <AccountIcon className={classes.arrowIcon} size={iconSize} />:{' '}
```
helfi92: Remove the column :)

PR #69, opened 2018-07-02
Filter bugs & issues by project

### LINKED ISSUE #59: Add filter based on the project
Add filter based on the project in the Toolbar on top of the task table

fiennyangeln:
Filter bugs & issues based on the project with default value as All project

Fix #59


src/utils/constants.js
```diff
@@ -5,3 +5,4 @@ export const ASSIGNEE = {
   UNASSIGNED: 'Unassigned',
   ASSIGNED: 'Assigned',
 };
+export const ALL = 'All';
```
helfi92: Add newline.

src/components/TasksTable/index.jsx
```diff
@@ -88,6 +106,9 @@ export default class TasksTable extends Component {
 
       return filteredItems
         .filter(item => !tag || item.tags.includes(tag))
+        .filter(
+          item => !project || project === 'All' || item.project === project
+        )
```
helfi92: Is there a way we can avoid looping on the list twice and perform both filters in one filter method?

src/components/TasksTable/index.jsx
```diff
@@ -220,6 +250,22 @@ export default class TasksTable extends Component {
...
                   ))}
                 </TextField>
+                <TextField
+                  select
+                  name="project"
+                  label="Project"
+                  value={project || ALL}
+                  className={classes.dropdown}
+                  onChange={this.handleFilterChange}>
+                  <MenuItem key={ALL} value={ALL}>
```
helfi92: Since we are not iterating over a list here, `key` is not required.

src/utils/constants.js
helfi92: `export const ALL_PROJECTS = 'All';`

PR #72, opened 2018-07-04
Make project card tab-able

### LINKED ISSUE #65: Move through projects using TAB
STR:

1. Navigate to http://codetribute.netlify.com/
2. Press Tab to have the cursor show inside the search bar
3. Press Tab again

Expected: The focus is now on the first project and the user is able to traverse the list of projects via the tabs key.
Actual: The focus goes somewhere besides the URL.

fiennyangeln:
Use tabIndex attribute and add style for focus

Fixes #65


fiennyangeln:
yay! thanks! 😃 @helfi92

PR #78, opened 2018-07-05
Create i'm feeling lucky button

### LINKED ISSUE #66: Create I'm Feeling Lucky button in bugs page
The button when clicked will redirect user to one of the bugs page

fiennyangeln:
The button will open new tab/window of the bug/issue page chosen
randomly with prioriy for unassigned bugs

Fix #66


src/components/TasksTable/index.jsx
```diff
@@ -143,6 +149,18 @@ export default class TasksTable extends Component {
     this.setQuery(newQuery);
   };
 
+  handleRandomButtonClick = () => {
+    const { items } = this.props;
+    const unassignedItems = items.filter(item => item.assignee === '-');
```
helfi92: We should create a utility function for unassigned and assigned filter functions since we are duplicating these functions now. Create `utils/assignment_filters`:

```jsx
export const unassigned = ({ assignee }) => assignee === '-';
export const assigned = ({ assignee }) => assignee !== '-';

```

src/components/TasksTable/index.jsx
```diff
@@ -155,77 +173,88 @@ export default class TasksTable extends Component {
...
-                  className={classes.dropdown}
-                  onChange={this.handleFilterChange}>
-                  {assignments.map(assignee => (
-                    <MenuItem key={assignee} value={assignee}>
-                      {assignee}
-                    </MenuItem>
+      <Fragment>
+        <div className={classes.flexContainer}>
+          <Button
+            variant="contained"
```
helfi92: The `contained` variant was introduced in [v1.2.0](https://github.com/mui-org/material-ui/releases/tag/v1.2.0) but `yarn.lock` is using `v.1.1.0` so this style is not being applied. If you want to keep the same style, you can just remove the `variant` prop and it should be okay.

fiennyangeln: ok! I change it into `raised` to keep the same style

src/components/TasksTable/index.jsx
```diff
@@ -143,6 +150,18 @@ export default class TasksTable extends Component {
     this.setQuery(newQuery);
   };
 
+  handleRandomButtonClick = () => {
```
helfi92: `handleRandomTaskClick`

src/components/TasksTable/index.jsx
```diff
@@ -155,77 +174,88 @@ export default class TasksTable extends Component {
...
-                  className={classes.dropdown}
-                  onChange={this.handleFilterChange}>
-                  {assignments.map(assignee => (
-                    <MenuItem key={assignee} value={assignee}>
-                      {assignee}
-                    </MenuItem>
+      <Fragment>
+        <div className={classes.flexContainer}>
+          <Button
+            variant="raised"
```
helfi92: `raised` puts too much emphasis on the button which we don't necessarily want to do here. Remove this line. We can use the default variant.

src/components/TasksTable/index.jsx
```diff
@@ -19,6 +20,7 @@ import classNames from 'classnames';
 import DataTable from '../DataTable';
 import sort from '../../utils/sort';
 import { ASSIGNEE } from '../../utils/constants';
+import { unassigned, assigned } from '../../utils/assignment_filters';
```
helfi92: My bad, JS file name convention is to use camel case. Rename to `assignmentFilters`.

PR #79, opened 2018-07-05
Fix error in the console

### LINKED ISSUE #76: Errors in the console when running the app locally
Running `codetribute` locally I get errors in the console. Let's make the console happy :)

In many of the material-ui components, the API exposes a `component` prop giving you the ability to change its DOM element.

screen shot 2018-07-04 at 10 10 22 am

fiennyangeln:
Use span in the typography component and use withRouter instead of
NavLink to navigate between pages

Fix #76

### DIFF
--- a/src/components/ProjectCard/index.jsx
+++ b/src/components/ProjectCard/index.jsx
@@ -1,13 +1,12 @@
 import { Component } from 'react';
 import { string, shape } from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
-import { withRouter } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 import Card from '@material-ui/core/Card';
 import CardContent from '@material-ui/core/CardContent';
 import Typography from '@material-ui/core/Typography';
 import Markdown from 'react-markdown';
 
-@withRouter
 @withStyles(theme => ({
   card: {
     textAlign: 'center',
@@ -27,7 +26,7 @@ import Markdown from 'react-markdown';
     fontWeight: 300,
     padding: 2 * theme.spacing.unit,
   },
-  navlink: {
+  link: {
     textDecoration: 'none',
   },
 }))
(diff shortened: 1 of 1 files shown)


src/components/ProjectCard/index.jsx
```diff
@@ -49,30 +50,49 @@ export default class ProjectCard extends Component {
...
+
   render() {
     const {
       classes,
-      project: { name, summary, fileName },
+      project: { name, summary },
     } = this.props;
 
     return (
-      <NavLink className={classes.navlink} to={`/projects/${fileName}`}>
```
helfi92: Let's use `Link` instead of `NavLink`.

```jsx
import { Link } from 'react-router-dom';`
```

src/components/ProjectCard/index.jsx
```diff
@@ -49,30 +50,49 @@ export default class ProjectCard extends Component {
...
+      const code = event.charCode || event.keyCode;
+
+      if (code !== 32 && code !== 13) {
+        return;
+      }
+    }
+
+    this.props.history.push(`/projects/${fileName}`);
+  };
+
```
helfi92: I would stay away from this if we can handle it using the `Link` component.

src/components/ProjectCard/index.jsx
fiennyangeln: @helfi92 but aren't both will render similarly as <a>?

src/components/ProjectCard/index.jsx
fiennyangeln: ok! will try it

src/components/ProjectCard/index.jsx
helfi92: `NavLink` is meant to be used for navigation. When the URL matches the `to` prop, it will attach an`active` class to the component which we don't want to do here. `NavLink` uses `Link` internally.

helfi92: I see. In that case, you can do `component="object"`. When an element is wrapped with `<object>` the DOM will consider this as an external resource and will be treated without looking at the parent element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object.

```jsx
<Typography component="object" ...>
   ...
</Typography>
```

fiennyangeln: WOW how did you find such simple solution ✨

PR #87, opened 2018-07-10
Feeling adventurous should open new tab while still giving random bugs on each click

### LINKED ISSUE #81: "I'm Feeling Adventurous" should open a new tab
Clicking on "I'm Feeling Adventurous" currently opens a new window. Let's have it open a new tab instead.

fiennyangeln:
Fix #81

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -180,7 +180,11 @@ export default class TasksTable extends Component {
       : items[Math.floor(Math.random() * items.length)].url;
 
     if (url) {
-      window.open(url, '_blank', 'noopener');
+      const thisWindow = window.open();
+
+      thisWindow.opener = null;
+      thisWindow.location = url;
+      thisWindow.target = '_blank';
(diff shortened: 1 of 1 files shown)


fiennyangeln:
@helfi92 im not sure if this is the best approach, another way i have thought of is having a button with rel & target properties, and href from the state and on each click it will change the state thus the href change. wdyt?

src/components/TasksTable/index.jsx
```diff
@@ -180,7 +180,11 @@ export default class TasksTable extends Component {
       : items[Math.floor(Math.random() * items.length)].url;
 
     if (url) {
-      window.open(url, '_blank', 'noopener');
+      const thisWindow = window.open();
+
+      thisWindow.opener = null;
+      thisWindow.location = url;
+      thisWindow.target = '_blank';
```
helfi92: We can probably do:

```js
Object.assign(window.open(), { opener: null, location: url, target: '_blank' });
```

PR #89, opened 2018-07-10
Fix Table Scrolling Issue

### LINKED ISSUE #88: Table scrolling issue
Scrolling the table horizontally scrolls "Bugs & Issues" and the filter icon as well. The expected outcome is to only have the table content be scrolled.

fiennyangeln:
Make the toolbar fixed when user scroll horizontally through the table
by moving the toolbar outside of the DataTable

Closes #88


src/components/TasksTable/index.jsx
```diff
@@ -214,9 +221,50 @@ export default class TasksTable extends Component {
     );
     const iconSize = 14;
     const projects = [...new Set(items.map(item => item.project))];
+    const filters = (
```
helfi92: No need to put this in a variable. You can simply inline it.

PR #104, opened 2018-07-16
Add skeleton for tests

helfi92:


helfi92:
At first, there was an error in both travis and Netlify. Re-running the travis build fixed it. I don't have a way to rerun the Netlify build but it looks like the yarn registry was acting silly for a bit. It should be okay.

PR #113, opened 2018-07-17
Show project description for bugzilla

fiennyangeln:
Fetch using bugzilla's graphql query when it is clicked

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -257,20 +257,14 @@ export default class TasksTable extends Component {
       async (name, onBugInfoClick) => {
         const item = this.props.items.find(item => item.summary === name);
 
-        if (item.url.indexOf('bugzilla') === -1) {
-          this.setState({
-            drawerOpen: true,
-            drawerItem: item,
-          });
-
-          return;
-        }
-
-        const description = await onBugInfoClick(item.id);
-
         this.setState({
           drawerOpen: true,
-          drawerItem: { ...item, description },
+          drawerItem: {
+            ...item,
+            ...(item.url.includes('bugzilla.mozilla.org')
+              ? { description: await onBugInfoClick(item.id) }
+              : null),
+          },
         });
       }
     )(name, this.props.onBugInfoClick);
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -1,4 +1,5 @@
 import { Component, Fragment } from 'react';
+import { ApolloConsumer } from 'react-apollo';
 import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import Chip from '@material-ui/core/Chip';
@@ -251,15 +252,30 @@ export default class TasksTable extends Component {
     this.setQuery({});
   };
 
-  handleDrawerOpen = ({ currentTarget: { name } }) => {
+  handleDrawerOpen = ({ currentTarget: { name } }, client) => {
     memoizeWith(
       name => name,
-      name =>
-        this.setState({
-          drawerOpen: true,
-          drawerItem: this.props.items.find(item => item.summary === name),
-        })
-    )(name);
+      (name, client, fetchComment) => {
+        const item = this.props.items.find(item => item.summary === name);
+
+        if (item.url.indexOf('bugzilla') === -1) {
+          this.setState({
+            drawerOpen: true,
+            drawerItem: item,
+          });
+
+          return;
+        }
+
+        fetchComment(client, item.id).then(description => {
+          item.description = description;
+          this.setState({
+            drawerOpen: true,
+            drawerItem: item,
+          });
+        });
+      }
+    )(name, client, this.props.fetchComment);
   };
 
   handleDrawerClose = () => {
@@ -354,12 +370,17 @@ export default class TasksTable extends Component {
                   {item.project}
                 </TableCell>
                 <TableCell className={classes.tableCell}>
-                  <IconButton
-                    name={item.summary}
-                    className={classes.infoButton}
-                    onClick={this.handleDrawerOpen}>
-                    <InformationVariantIcon />
-                  </IconButton>
+                  <ApolloConsumer>
+                    {client => (
+                      <IconButton
+                        name={item.summary}
+                        client={client}
+                        className={classes.infoButton}
+                        onClick={e => this.handleDrawerOpen(e, client)}>
+                        <InformationVariantIcon />
+                      </IconButton>
+                    )}
+                  </ApolloConsumer>
                   <List dense disablePadding className={classes.summary}>
                     <ListItem
                       classes={{
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -1,5 +1,4 @@
 import { Component, Fragment } from 'react';
-import { ApolloConsumer } from 'react-apollo';
 import TableCell from '@material-ui/core/TableCell';
 import TableRow from '@material-ui/core/TableRow';
 import Chip from '@material-ui/core/Chip';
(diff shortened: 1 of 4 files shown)


src/components/TasksTable/index.jsx
```diff
@@ -347,12 +363,17 @@ export default class TasksTable extends Component {
...
-                    name={item.summary}
-                    className={classes.infoButton}
-                    onClick={this.handleDrawerOpen}>
-                    <InformationVariantIcon />
-                  </IconButton>
+                  <ApolloConsumer>
+                    {client => (
+                      <IconButton
+                        name={item.summary}
+                        client={client}
```
helfi92: Why are we passing `client` to `IconButton`?

src/components/TasksTable/index.jsx
```diff
@@ -245,15 +246,30 @@ export default class TasksTable extends Component {
     this.setQuery({});
   };
 
-  handleDrawerOpen = ({ target: { name } }) => {
+  handleDrawerOpen = ({ target: { name } }, client) => {
```
helfi92: Why are we passing `client` here?

src/components/TasksTable/index.jsx
```diff
@@ -245,15 +246,30 @@ export default class TasksTable extends Component {
...
+        if (item.url.indexOf('bugzilla') === -1) {
+          this.setState({
+            drawerOpen: true,
+            drawerItem: item,
+          });
+
+          return;
+        }
+
+        fetchComment(client, item.id).then(description => {
```
helfi92: Use await/async instead.

src/components/TasksTable/index.jsx
fiennyangeln: because we need it to do client.query?

helfi92: Can you do `this.props.client.query()` inside `src/views/Project/index.jsx` instead?

src/views/Project/index.jsx
```diff
@@ -155,6 +156,23 @@ export default class Project extends Component {
       this.load();
     }
   }
+
+  handleFetchBugFirstComment = async id => {
+    try {
+      const {
+        data: { comments },
+      } = await this.props.client.query({
```
helfi92: I think you can `import { withApollo } from 'react-apollo';` and use @withApollo on top of the class to have access to `this.props.client`. You can then remove `<ApolloConsumer>` from `src/App.jsx`.

src/views/Project/index.jsx
```diff
@@ -155,6 +156,23 @@ export default class Project extends Component {
       this.load();
     }
   }
+
+  handleFetchBugFirstComment = async id => {
```
helfi92: Sorry, I think `handleBugInfoClick` makes more sense here.

src/views/Project/index.jsx
```diff
@@ -155,6 +156,23 @@ export default class Project extends Component {
...
+        data: { comments },
+      } = await this.props.client.query({
+        query: commentsQuery,
+        variables: { id },
+        context: { link: 'bugzilla' },
+      });
+
+      return comments[0].text;
+    } catch (err) {
+      return '';
```
helfi92: This doesn't seem to do anything with the error. If anything goes wrong, we can handle it using `componentDidCatch`. In tc-web, it's done in https://github.com/taskcluster/taskcluster-web/blob/2998f6561a4de26908d4a2eba4aaca8443c15a53/src/App/index.jsx#L79-L81. Once the error state is set, `App/index.jsx` displays the ErrorPanel.

src/components/TasksTable/index.jsx
```diff
@@ -248,12 +248,26 @@ export default class TasksTable extends Component {
   handleDrawerOpen = ({ target: { name } }) => {
     memoizeWith(
       name => name,
-      name =>
+      async (name, onBugInfoClick) => {
+        const item = this.props.items.find(item => item.summary === name);
+
+        if (item.url.indexOf('bugzilla') === -1) {
```
helfi92: This will not work if the github url contains the the string bugzilla (e.g., https://github.com/mozilla/bugzilla). You can do:

```js
item.url.includes('bugzilla.mozilla.org');
```

src/components/TasksTable/index.jsx
```diff
@@ -248,12 +248,26 @@ export default class TasksTable extends Component {
...
+            drawerOpen: true,
+            drawerItem: item,
+          });
+
+          return;
+        }
+
+        const description = await onBugInfoClick(item.id);
+
         this.setState({
```
helfi92: To remove duplication, I think you can do:

```jsx
this.setState({
  drawerOpen: true,
  drawerItem: item,
  ...(item.url.includes('bugzilla.mozilla.org') ? { description: await onBugInfoClick(item.id) } : null), 
});

```

src/views/Project/index.jsx
fiennyangeln: I tried putting `componentDidCatch` in the Project/index.jsx then i change the query comments to be commentzz which will return error (ApolloError), but it only show the error in the console instead of the panel, then I move it to the App/index.jsx and it still show only in the console. is it because the error came from event handler? https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers 

i agree the catch block here did nothing so i added setState in the catch section

helfi92: Did you make sure to remove the try and catch statement before running the query again?

fiennyangeln: yes

src/views/Project/index.jsx
```diff
@@ -158,6 +161,23 @@ export default class Project extends Component {
...
   }
+
+  handleBugInfoClick = async id => {
+    try {
+      const {
+        data: { comments },
+      } = await this.props.client.query({
+        query: commentsQuery,
+        variables: { id },
+        context: { link: 'bugzilla' },
```
helfi92: Where is the `context` used?

fiennyangeln: context is to choose between bugzilla / github link afaik

helfi92: Could you point me to where `context` is being used for the comment?

helfi92: Could you point me to where this is being used for the comment?

fiennyangeln: https://github.com/mozilla-frontend-infra/codetribute/blob/master/src/App.jsx#L35 we use the split method which will choose what link to send request to according to the context, 

but now that i see the condition, since we check for github otherwise set it to bugzilla, maybe we can remove it in here?

helfi92: Oh, shouldn't this be `context: { client: 'bugzilla' },` instead?

fiennyangeln: OHHH so thats what u mean!! 🤦‍♀️

src/components/TasksTable/index.jsx
```diff
@@ -253,12 +253,20 @@ export default class TasksTable extends Component {
   handleDrawerOpen = ({ target: { name } }) => {
```
helfi92: When we use `memoizeWith`, we are caching a value (the return value). The async function is doing setState but not returning any value so we are not really caching any value. Also, we should not be setting state inside `memoizeWith` because that mean this state will not be set when the value is already cached.

src/views/Project/index.jsx
```diff
@@ -158,6 +161,23 @@ export default class Project extends Component {
...
+    try {
+      const {
+        data: { comments },
+      } = await this.props.client.query({
+        query: commentsQuery,
+        variables: { id },
+        context: { client: 'bugzilla' },
+      });
+
+      return comments[0].text;
```
helfi92: I would probably cache this function instead.

src/views/Project/index.jsx
```diff
@@ -159,6 +162,27 @@ export default class Project extends Component {
       this.load();
     }
   }
+
+  handleBugInfoClick = id =>
```
helfi92: `handleBugInfoClick = memoizeWith(...);`

src/views/Project/index.jsx
```diff
@@ -159,6 +162,27 @@ export default class Project extends Component {
...
+            variables: { id },
+            context: { client: 'bugzilla' },
+          });
+
+          return comments[0].text;
+        } catch (error) {
+          this.setState({ error });
+        }
+      }
+    )(id);
```
helfi92: By returning the memoizeWith function, you won't need to do `(id)`.

PR #127, opened 2018-07-20
Enable mentored bugs from bugzilla to be shown

### LINKED ISSUE #32: Handle mentored bugs for bugzilla
As of now, the bugs we are fetching from bugzilla are those labeled 'good-first-bug' as its tag, when I check on a few website linking to bugsahoy, most of the site mentioned that BugsAhoy are meant for mentored bugs so i think we may need to add capabilities for mentored bugs
TODO: find appropriate query to query mentored bugs

fiennyangeln:
-Refactor project/index.jsx
-Separate it into two routes, /projects/:project/mentored and
/projects/:project
-Move the AppBar into the project/index.jsx since it is similar
for both route
-Add mentored bugs filter using f1 and o1 properties

Closes #32


helfi92:
How do I get this running? I am running https://github.com/mozilla-frontend-infra/bugzilla-graphql-gateway/pull/8 locally for the backend but when I click the "Show me mentored bugs/issues" button I get a network error.

screen shot 2018-07-23 at 10 52 31 am

fiennyangeln:
@helfi92 Sorry forgot to write it down. Use this branch https://github.com/mozilla-frontend-infra/bugzilla-graphql-gateway/pull/8

helfi92:
Are we able to do a query for mentored bugs in addition to the normal bugzilla query we usually do in the beginning with keyword good-first-bug? This would be a better experience than having the user click the filter icon then click the show mentored. I'm okay having to do two separate queries.

helfi92:
What do you mean by "all the bugzilla query"?

fiennyangeln:
i mean including the fetchMore we do after the initial query

helfi92:
Sounds good. Let's give it a try and see how it goes.

fiennyangeln:
@helfi92 found a way not to do more fetch, wdyt?

src/components/TasksTable/index.jsx
```diff
@@ -335,11 +342,26 @@ export default class TasksTable extends Component {
...
+              component={Link}
+              to={{
+                pathname: url.includes('mentored')
+                  ? url.replace('mentored', '')
+                  : `${url}/mentored`.replace('//', '/'),
+                search: this.props.location.search,
+              }}>
+              SHOW{url.includes('mentored') ? ' GOOD FIRST ' : ' MENTORED '}
+              BUGS / ISSUES
+            </Button>
```
helfi92: I believe this button can be removed now.

src/views/Project/bugs.graphql
```diff
@@ -1,21 +1,24 @@
...
+      component
+      lastChanged
+      id
+      keywords
+      summary
+      whiteboard
     }
   }
-}
+}
```
helfi92: Add newline.

src/views/Project/index.jsx
```diff
@@ -92,8 +92,23 @@ const tagReposMapping = repositories =>
     }) => ({
       fetchPolicy: 'network-only',
       variables: {
-        search: {
+        searchGoodFirst: {
```
helfi92: Rename to `goodFirst`.

src/views/Project/index.jsx
```diff
@@ -92,8 +92,23 @@ const tagReposMapping = repositories =>
...
+          ...(productsWithNoComponents(projects[project].products).length
+            ? { products: productsWithNoComponents(projects[project].products) }
+            : {
+                // otherwise, get only the first product and its components
+                // as component is not unique for product thus can't be merged
+                products: Object.keys(projects[project].products[0])[0],
+                components: Object.values(projects[project].products[0])[0],
+              }),
+        },
+        searchMentored: {
```
helfi92: Rename to `mentored`.

src/components/TasksTable/index.jsx
```diff
@@ -335,11 +342,26 @@ export default class TasksTable extends Component {
               ))}
             </TextField>
             <Button
+              className={classes.spacing}
```
helfi92: I think we can remove this class.

PR #132, opened 2018-07-23
Fix drawer open in chrome and mobile

### LINKED ISSUE #130: Drawer not showing anything besides the x icon on mobile
drawer in mobile

fiennyangeln:
Use currentTarget instead of target : based on this https://github.com/mui-org/material-ui/issues/7974

Fix #130

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -251,7 +251,7 @@ export default class TasksTable extends Component {
     this.setQuery({});
   };
 
-  handleDrawerOpen = ({ target: { name } }) => {
+  handleDrawerOpen = ({ currentTarget: { name } }) => {
     memoizeWith(
       name => name,
       name =>


src/components/TasksTable/index.jsx
```diff
@@ -251,13 +251,15 @@ export default class TasksTable extends Component {
...
-  handleDrawerOpen = ({ target: { name } }) => {
+  handleDrawerOpen = ({ currentTarget: { name } }) => {
     memoizeWith(
       name => name,
       name =>
         this.setState({
           drawerOpen: true,
-          drawerItem: this.props.items.find(item => item.summary === name),
+          drawerItem: this.props.items.find(
+            item => sort(item.summary, name) === 0
```
helfi92: What's wrong with `item.summary === name`?

fiennyangeln: it fails the travis build. i tried it 2 times.
At first, i wonder why the travis fail, then i randomly change the find using sort method, so that i can force to rerun the travis (i remember when we add testing it also fail at first then you rerun it somehow and it works) , then, it runs successfully, but then i do

`g reset --hard HEAD^` 
`g push origin fix/handle-drawer-open-on-chrome-and-mobile -f`

to remove the last commit about the find using sort then it fail again

helfi92: Can you try again (remove `sort`) and send me the link to the failed travis build, please? I'll take a look.

fiennyangeln: ok removed! this is the link https://travis-ci.com/mozilla-frontend-infra/codetribute/builds/79747509

helfi92: Seems to be okay now.

PR #133, opened 2018-07-23
Add Rust projects

Manishearth:
May also be worth linking to https://www.rustaceans.org/findwork

Is there a way I can mark these as Rust projects so that folks can filter by Rust to find bugs?

### DIFF
--- a/src/data/cargo.yaml
+++ b/src/data/cargo.yaml
@@ -1,4 +1,4 @@
-name: cargo
+name: Cargo
 summary: Cargo is Rust's build system
 repositories: 
  - rust-lang-nursery/rustfmt: E-Easy
--- a/src/data/cargo.yaml
+++ b/src/data/cargo.yaml
@@ -1,0 +1,4 @@
+name: cargo
+summary: Cargo is Rust's build system
+repositories: 
+ - rust-lang-nursery/rustfmt: E-Easy
--- a/src/data/clippy.yaml
+++ b/src/data/clippy.yaml
@@ -1,4 +1,4 @@
-name: clippy
+name: Clippy
 summary: Clippy is linter for Rust code
 repositories: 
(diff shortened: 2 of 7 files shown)


djmitche:
OK, sounds good.  I don't really like the granularity of Firefox either..

Manishearth:
Yeah, but this is a way less granular thing, and like I said these projects are pretty different.

fiennyangeln:
You can add Rust before the project name thus it appears close to each other and then it can be filtered later using the search box.

Manishearth:
Done.

Though, more in general, I was asking for a language-based search like bugsahoy and https://starters.servo.org/ , where the tool understands how different projects tag languages (and understands some projects as "default language X", so you can search what you're familiar with.

As in, not just "I want to work on a project that's a part of Rust" (accomplished by searching for "rust"), but "I want to work on a project that's written in rust" (like servo)

fiennyangeln:
Oh ok understand it! like this is it ? https://www.joshmatthews.net/bugsahoy/?rust=1

Manishearth:
yep!

src/data/cargo.yaml
```diff
@@ -0,0 +1,4 @@
+name: cargo
```
fiennyangeln: can we use pascal case here (first letter capitalized) to make it look similar to other project?

src/data/clippy.yaml
```diff
@@ -0,0 +1,4 @@
+name: clippy
```
fiennyangeln: can we use pascal case here (first letter capitalized) to make it look similar to other project?

src/data/rustfmt.yaml
```diff
@@ -0,0 +1,4 @@
+name: rustfmt
```
fiennyangeln: can we use pascal case here (first letter capitalized) to make it look similar to other project?

PR #151, opened 2018-07-26
Improve SEO and Accessibility score on Lighthouse

fiennyangeln:
-Add aria-label to input and button
-Fix Lists do not contain only `<li>` elements and script supporting
-Fix document does not have meta description

### DIFF
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -2,8 +2,14 @@ module.exports = {
   use: [
     ['neutrino-preset-mozilla-frontend-infra/react', {
       html: {
-        title: 'Codetribute'
-      }
+        title: 'Codetribute',
+        meta: [
+          {
+            name: 'Description',
+            content: 'A web app to help people make their first code contribution with Mozilla',
+          },
+        ],
+      },
     }],
     ['@neutrinojs/env', ['GITHUB_PERSONAL_API_TOKEN', 'BUGZILLA_ENDPOINT', 'NODE_ENV']],
     (neutrino) => {
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -6,7 +6,7 @@ module.exports = {
         meta: [
           {
             name: 'Description',
-            content: 'A web app to help people make their first code contribution with Mozilla',
+            content: 'Find your first code contribution with Mozilla',
           },
         ],
       },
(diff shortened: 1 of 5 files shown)


.neutrinorc.js
```diff
@@ -2,8 +2,14 @@ module.exports = {
   use: [
     ['neutrino-preset-mozilla-frontend-infra/react', {
       html: {
-        title: 'Codetribute'
-      }
+        title: 'Codetribute',
+        meta: [
+          {
+            name: 'Description',
+            content: 'A web app to help people make their first code contribution with Mozilla',
```
helfi92: `content: 'Find your first code contribution with Mozilla',`

src/components/TasksTable/index.jsx
```diff
@@ -295,7 +295,7 @@ export default class TasksTable extends Component {
     return (
       <Fragment>
         <Toolbar className={classes.toolbar}>
-          <Typography variant="title" className={classes.title}>
+          <Typography variant="title" className={classes.title} id="tableTitle">
```
helfi92: Remove `id`.

src/components/TasksTable/index.jsx
```diff
@@ -305,7 +305,7 @@ export default class TasksTable extends Component {
             className={classes.adventurousButton}>
             I’m Feeling Adventurous
           </Button>
-          <IconButton onClick={this.handleFilterToggle}>
+          <IconButton aria-label="Filter" onClick={this.handleFilterToggle}>
```
helfi92: Nice!

src/components/TasksTable/index.jsx
```diff
@@ -360,11 +360,16 @@ export default class TasksTable extends Component {
                 <TableCell className={classes.tableCell}>
                   <IconButton
                     name={item.summary}
+                    aria-label="Info"
```
helfi92: Rename to `Information`.

src/components/TasksTable/index.jsx
fiennyangeln: I add id because the table is rendered as `<table aria-labelledby='tableTitle'>`

helfi92: Ahh, ok makes sense. Thanks for clarifying.

PR #154, opened 2018-07-26
Improve Codetribute's observatory scan results

helfi92:
https://observatory.mozilla.org/analyze/codetribute.netlify.com

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,7 +5,33 @@
 [build.environment]
   YARN_FLAGS = "--frozen-lockfile"
 
+# Redirect HTTP to HTTPS
+# TODO: Delete when Codetribute successfully moves to codetribute.mozilla.org
+[[redirects]]
+  from = "http://codetribute.netlify.com/*"
+  to = "https://www.codetribute.netlify.com/:splat"
+  status = 301
+  force = true
+
+# Redirect HTTP to HTTPS
+[[redirects]]
+  from = "http://codetribute.mozilla.org/*"
+  to = "https://www.codetribute.mozilla.org/:splat"
+  status = 301
+  force = true
+
+# Rule for SPA
 [[redirects]]
   from = "/*"
   to = "/index.html"
   status = 200
+
+[[headers]]
+  for = "/*"
+  [headers.values]
+    X-Frame-Options = "DENY"
+    X-XSS-Protection = "1; mode=block"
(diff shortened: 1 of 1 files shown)


eliperelman:
Current results: https://observatory.mozilla.org/analyze/deploy-preview-154--codetribute.netlify.com

netlify.toml
```diff
@@ -5,7 +5,33 @@
 [build.environment]
   YARN_FLAGS = "--frozen-lockfile"
 
+# Redirect HTTP to HTTPS
+# TODO: Delete when Codetribute successfully moves to codetribute.mozilla.org
+[[redirects]]
+  from = "http://codetribute.netlify.com/*"
+  to = "https://www.codetribute.netlify.com/:splat"
```
eliperelman: Why is there a `www` here?

helfi92:
I believe the redirection will not work on `https://deploy-preview-154--codetribute.netlify.com/` because it will fail the `from` condition.

eliperelman:
Does netlify support wildcarding parts of the URL too?

netlify.toml
eliperelman: Right, but that only applies to top-level domains, e.g. `mozilla.com`. It would not apply to `codetribute.mozilla.com`. You can remove the `www` from the redirect.

helfi92:
> Does netlify support wildcarding parts of the URL too?

Netlify allows the usage of [splats](https://www.netlify.com/docs/redirects/#splats) and [placeholders](https://www.netlify.com/docs/redirects/#placeholders).

netlify.toml
helfi92: Done.

djmitche:
looking good!

netlify.toml
```diff
@@ -5,7 +5,33 @@
...
   from = "/*"
   to = "/index.html"
   status = 200
+
+[[headers]]
+  for = "/*"
+  [headers.values]
+    X-Frame-Options = "DENY"
+    X-XSS-Protection = "1; mode=block"
+    Content-Security-Policy = "frame-ancestors 'self'"
```
djmitche: It looks like Observatory wanted more?  Were those problematic?

djmitche:
https://media.giphy.com/media/ipweLwfKEsaKk/giphy.gif

PR #157, opened 2018-07-27
UI for getting bug with language filter without project

fiennyangeln:


helfi92:
> "Feature/ui for bug without project"

What feature is that?

fiennyangeln:
@helfi92 UI for going to the bug page with language filter without project

fiennyangeln:
sorry  i just use my branch name directly! but idk if this UI is good as initially we dont have language selected, thus no query at first?

helfi92:
Should we have a skill column similar to https://contribute.testpilot.firefox.com/tasks/ where we can a list of skills?

fiennyangeln:
should we place it in a drawer so that we can stilll use the tasktable? yeah i think we can add the language icon too i just checked the mdi icons

helfi92:
Yeah, a drawer would be good :)

src/views/Languages/index.jsx
```diff
@@ -96,22 +140,25 @@ export default class Languages extends Component {
...
+          open={drawerOpen}
+          onClose={this.handleDrawerToggle}
+          classes={{
+            paper: classes.drawerPaper,
+          }}
+          ModalProps={{
+            keepMounted: true,
+          }}>
+          {drawer}
+        </Drawer>
```
helfi92: Picking a skill shows an empty table. Is this on purpose?

screen shot 2018-07-30 at 10 10 28 am

fiennyangeln: yes it is supposed to be handled by other PR #163

PR #158, opened 2018-07-27
Update addons.yaml

muffinresearch:


src/data/addons.yaml
```diff
@@ -1,3 +1,45 @@
...
+    to improve their add-ons and create new ones when they know people 
+    appreciate their work.
+  * Tell your friends and family that Firefox is a fast, secure browser that 
+    protects their privacy, and they can use add-ons to make it their own!
+
+  To see more ways you can contribute to the add-on community, please visit 
+  our [wiki].
+
+  [Make your own add-on]: https://addons.mozilla.org/developers/
+  [good first bug]: https://github.com/search?l=&q=repo:mozilla/addons+repo:mozilla/addons-frontend+repo:mozilla/addons-linter+repo:mozilla/addons-server+label:%22contrib:+good+first+bug%22&ref=advsearch&state=open&type=Issues
```
fiennyangeln: you can add github repo too here! 
the format is like this (https://github.com/mozilla-frontend-infra/codetribute#github) :
so based on your link, it would be like this:
```
repositories:
 - mozilla/addons: "contrib: good first bug"
 - mozilla/addons-frontend: "contrib: good first bug"
 - mozilla/addons-linter: "contrib: good first bug"
 - mozilla/addons-server: "contrib: good first bug"
```

muffinresearch: I've add the repos to this patch.

fiennyangeln: Thanks!

PR #167, opened 2018-07-31
Update CSP

helfi92:


helfi92:
You should be able to see that the site works once the Netlify build finishes.

PR #163, opened 2018-07-30
Fetch bugzilla bugs with language query without project

fiennyangeln:
This is branched from #157. 

- Map the bugzilla whiteboard query name and standard name (in utils.js)
- Query from bugzilla using the whiteboard tag
- Display the query result and the possible error

Now that we don't do more fetchMore should I add loading state?

Closes #155

Issue #156, opened 2018-07-27
Fetch bug from github with only language specified


fiennyangeln:
- Fetch the list of language available then fetch the bug for that repository

PR #169, opened 2018-08-01
Fetch github issues from selecting languages without specifying project

fiennyangeln:
- This PR is branched from #163 
- The commits related to this change are the ones started with `Add fragmentTypes to handle warning heuristic m.......` and commit after it
- Solves #156


helfi92:
I will hold on to the review until we settle on https://github.com/mozilla-frontend-infra/codetribute/pull/163.

helfi92:
This should now be in a state where you can implement the changes we agreed on in https://github.com/mozilla-frontend-infra/codetribute/pull/163. Feel free to either rebase your work or start a new PR with a fresh branch.

fiennyangeln:
Ok! rebased to master 😅

src/views/Languages/index.jsx
```diff
@@ -1,69 +1,266 @@
...
+        },
       },
+      context: {
+        client: 'bugzilla',
+      },
+    }),
+  })
+)
+@withStyles(theme => ({
+  root: {
```
helfi92: There seems to be a lot of unused styles in this file. Probably from the rebase.

src/fragmentTypes.json
```diff
@@ -0,0 +1,30 @@
+{
```
helfi92: Could you explain the use of this file, please?

src/components/Sidebar/index.jsx
```diff
@@ -51,7 +51,7 @@ export default class Sidebar extends Component {
 
     return (
       <List disablePadding>
-        {Object.keys(BUGZILLA_LANGUAGES).map(language => (
+        {['Swift', ...Object.keys(BUGZILLA_LANGUAGES)].map(language => (
```
helfi92: Why not just include `Swift` in BUGZILLA_LANGUAGES?

fiennyangeln: it does not seem to have the mapping in the whiteboards

src/fragmentTypes.json
fiennyangeln: so since at first we fetch for repository then after that for query, the data shape is different. If we didnt use this, only the field that is inside the first query will be saved and passed as props. and this warning appear `WARNING heuristic fragment matching going on`. so to enable the `updateMore` to merge the previous and more result including all its field, we need the IntrospectionFragmentMatcher to tell that there is 'union' of different data field

src/fragmentTypes.json
```diff
@@ -0,0 +1,30 @@
...
+          {
+            "name": "User"
+          },
+          {
+            "name": "Organization"
+          },
+          {
+            "name": "MarketplaceListing"
+          }
+        ]
```
helfi92: I'm trying to understand this file. Why do we have some many possible types listed here when we don't care about `MarketplaceListing`, `User`, etc.. Maybe if you can link me to some documentation online, it will help me understand.

src/utils/constants.js
```diff
@@ -39,3 +39,4 @@ export const BUGZILLA_LANGUAGES = {
   XML: 'xml',
   XUL: 'xul',
 };
+export const GITHUB_LANGUAGE_QUERY_ORDER = { field: 'SIZE', direction: 'DESC' };
```
helfi92: What does `field: 'SIZE'` mean? Can you show me a link online to where this is explained, please?

src/views/issues.graphql
```diff
@@ -1,5 +1,5 @@
-query Issues($searchQuery: String!) {
-  search(first: 100, type: ISSUE, query: $searchQuery) {
+query Issues($searchQuery: String!, $type: SearchType = ISSUE, $order:LanguageOrder) {
```
helfi92: Add a space after `$order:`.

src/views/issues.graphql
```diff
@@ -20,6 +20,14 @@ query Issues($searchQuery: String!) {
           }
         }
       }
+      ...on Repository {
+        nameWithOwner
+        languages(first:1, orderBy:$order) {
```
helfi92: `first: 1, orderBy: $order`. Also I think we should rename order to orderBy.

src/components/Sidebar/index.jsx
helfi92: Ah, in that case, add a comment to make it more clear for someone else reading the code.

src/utils/constants.js
fiennyangeln: https://developer.github.com/v4/object/repository/ go to the languages section, it basically says that it will return the languages a repository has based on its size (starting from the major ones to the minor ones like html). and we use this later when we add field `first: 1` to be only showing the first (only show the most used language)

src/fragmentTypes.json
fiennyangeln: I do it based on this guide. https://www.apollographql.com/docs/react/advanced/fragments.html#fragment-matcher but initially it was 1000+ lines, then I start reducing it to only have UNION that we use.

src/utils/constants.js
helfi92: Makes sense. Thanks for the explaining.

helfi92: @fiennyangeln There seems to be a `primaryLanguage` field. Shouldn't we use this instead of doing `first: 1`? https://developer.github.com/v4/object/repository/#fields

src/views/issues.graphql
```diff
@@ -1,5 +1,5 @@
-query Issues($searchQuery: String!) {
-  search(first: 100, type: ISSUE, query: $searchQuery) {
+query Issues($searchQuery: String!, $type: SearchType = ISSUE) {
```
helfi92: Do we still need https://github.com/mozilla-frontend-infra/codetribute/pull/169/files#diff-fd6b3e26da5d99c15109f388dc828f31R7 now that we narrowed down the SearchType to ISSUE?

src/components/Sidebar/index.jsx
```diff
@@ -49,9 +49,10 @@ export default class Sidebar extends Component {
       CSS: <LanguageCss3Icon />,
     };
 
+    // Add languages that does not have BUGZILLA MAPPING
```
helfi92: I'm not sure I understand this comment. Can you elaborate? Also, when I said to add a comment, I meant explaining why we are doing `['Swift', ...Object.keys(BUGZILLA_LANGUAGES)]` instead of just including `Swift` in `BUGZILLA_LANGUAGES`.

src/views/issues.graphql
fiennyangeln: i think we still need it when there is ... in the query. we still query for Repository first, its just that I set the searchType default value to ISSUE. Maybe i need a better naming for the query to indicate it

fiennyangeln:
ok! do you think i need to change the github query name too?

helfi92:
What do you want to rename it to?

fiennyangeln:
githubInfo?

helfi92:
Sure.

helfi92:
hmm, wouldn't `githubIssues` make more sense?

fiennyangeln:
but doesnt github issues mean only issues not repository?

helfi92:
Right. nevermind my suggestion then.

helfi92:
Don't forget to click "Update branch" before squashing and merging.

PR #188, opened 2018-08-08
Reduce horizontal space by reducing the padding right

### LINKED ISSUE #187: Make the project column take less space
The project column has a lot of unused white space. Let's find a way to make the column take less.

screen shot 2018-08-08 at 11 37 42 am

We can use the "..." when the project name is long and `title` prop for the tooltip similar to the summary column.

fiennyangeln:
Make the paddingRight similar to paddingLeft

Closes #187


src/theme.jsx
```diff
@@ -77,6 +77,7 @@ const theme = createMuiTheme({
     },
     MuiTableCell: {
       root: {
+        paddingRight: 24,
```
helfi92: Replace with `padding: '4px 16px'`.

PR #189, opened 2018-08-09
Add logo as button to the homepage

### LINKED ISSUE #177: Provide a way to go to the homepage from nested views
If a user is on the project or languages page, there is no easy way to navigate to the homepage. I see two options:
* Add a back button somewhere
* Add "Contribute" somewhere in the app bar and have it be clickable

fiennyangeln:
Add icon so that it can be reusable later for language icon too

Closes #177


fiennyangeln:
how about home icon?

helfi92:
No home icon anymore, let's use the back button. We will still keep the logo for the favicon, however :)

PR #226, opened 2018-09-19
Add icon to project card

vaishnavipamulapati:
I've used the material design icons, svg files and also added a default icon (the same one used for the languages on the sidebar). Should we use the default icon in place of the svg files because of color? I've included mdi css because I wasn't sure how to dynamically import the right component from mdi-react.

In the project YAML files, icon or svgUrl can be specified. If not, the default icon 'web' will be used. 
codetribute_icon_screenshot


vaishnavipamulapati:
The svgs and icons are failing to load due to a "Content Security Policy directive" on the netlify deployment. Should I be loading both of them locally?

helfi92:
This is awesome! I will look at it tomorrow :)

.neutrinorc.js
```diff
@@ -16,6 +16,10 @@ module.exports = {
           {
             href: './static/favicon.png',
             rel: 'shortcut icon',
+          },
+          {
+            href: 'https://cdn.materialdesignicons.com/2.7.94/css/materialdesignicons.min.css',
```
fiennyangeln: I feel like there is no need to add materialdesignicons CDN, afaik, we had [mdi-react](https://github.com/levrik/mdi-react) library from which you can add all the icons using something like `<TestTubeIcon />`

src/components/ProjectCard/index.jsx
```diff
@@ -59,13 +67,32 @@ export default class ProjectCard extends Component {
...
+              <i
+                className={`mdi ${icon}`}
+                style={{ fontSize: '50px', margin: '10px' }}
+              />
+            )}
+            {svgUrl && (
+              <img
+                alt="svgIcon"
+                src={svgUrl}
+                style={{ width: '50px', height: '50px', fill: '#4bacb8' }}
```
fiennyangeln: Can you try moving the style to above (inside the `withStyles`)?

.neutrinorc.js
helfi92: Agreed. We can specify the icons in the yaml file and dynamically import the icons. I will try to send a patch for that :)

vaishnavipamulapati:
This is great, thanks! I didn't know about the import function and hence added the css. 
Is it fine that I added the default icon in case there's no suitable icon present in mdi-react or there isn't an svg icon available? (For example: Servo)

helfi92:
Oh yeah, that's fine :) We should, however, try to add an icon for servo. @vaishnavipamulapati Do you want to add one?

jdm:
https://github.com/servo/servo/blob/master/resources/servo.png is one option, or https://github.com/servo/servo/blob/master/resources/Servo.ico, or https://github.com/servo/servo/blob/master/resources/servo.svg.

vaishnavipamulapati:
Oh, I didn't find the svg icon for it then. Thank you @jdm for sending the links! 
I have added the svg icon for servo now :D

README.md
```diff
@@ -116,6 +117,13 @@ repositories:
 ```
 _Note: The `summary` and `introduction` fields are to be rendered as markdown, allowing for bullet points, links and other simple formatting._
 
+#### Project Icons
+
+Each project should have an icon associated to it to be displayed on the home page. To pick an icon, you could either:
+1. Pick an icon from [materialdesignicons](https://materialdesignicons.com) and include its name in the `icon` field of the yaml file.
+2. Include an icon in `src/images/projectIcons`. The icon should have the same name as its configuration file
+and have the image extension set as `.svg`.
```
djmitche: In this case, should I not include `icon:` in the yaml?

helfi92: Correct.

helfi92:
🎉

PR #233, opened 2018-09-24
Remove data URI icons

helfi92:
The rust project icon was not being displayed because the CSP
don't allow data URIs.


helfi92:
It means hey, webpack, don't have these images loaded as data-URIs when the byte size is bigger than 1. https://neutrinojs.org/packages/image-loader/#usage.

fiennyangeln:
@helfi92 thanks 😄

PR #240, opened 2018-10-04
Update yaml loader without breaking the build

fiennyangeln:


.neutrinorc.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
         .rule('js-yaml')
           .test(/\.(yaml|yml)$/)
           .use('js-yaml-loader')
+          .options({safe: true})
```
helfi92: Can you point me where I can find more information about this option?

helfi92: Clicking on the link shows:
screen shot 2018-10-04 at 11 28 56 am

fiennyangeln: @helfi92 ah sorry! I tried pointing it to the README file but it ends up not to be a link x_x here's the link https://github.com/wwilsman/js-yaml-loader/compare/v0.1.0...v1.0.0 then go to the readme file part

helfi92: `.neutrinorc.js` is not linted since it's not part of the `src` directory. Could we add some spacing:

```js
.options({ safe: true })
```

helfi92: Thank you!

fiennyangeln: sure!

PR #243, opened 2018-10-05
Horizontal scrollbar when viewing projects

### LINKED ISSUE #242: Horyzonal scrollbar on 1920x1080
immagine
testing with devtools seems that if I change max-width to `calc(100vw - 260px);` but is not clear to me how to patch.
Anyway this fix my issue :-)

Mte90:
Fixes #242.

### DIFF
--- a/src/views/Projects/index.jsx
+++ b/src/views/Projects/index.jsx
@@ -34,7 +34,7 @@ import sort from '../../utils/sort';
     margin: '0 auto',
     [theme.breakpoints.up('md')]: {
       padding: ' 0px 12px',
-      maxWidth: `calc(100vw - ${theme.drawerWidth}px)`,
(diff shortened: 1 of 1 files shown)


src/views/Projects/index.jsx
```diff
@@ -34,7 +34,7 @@ import sort from '../../utils/sort';
     margin: '0 auto',
     [theme.breakpoints.up('md')]: {
       padding: ' 0px 12px',
-      maxWidth: `calc(100vw - ${theme.drawerWidth}px)`,
+      maxWidth: `calc(100vw - ${20 + theme.drawerWidth}px)`,
```
helfi92: I think something like `maxWidth: '100%'` might fix it. Give it a try.

Mte90:
confirmed works also with 100%

PR #247, opened 2018-10-08
Update webextensions.yaml

Rob--W:
The BMO categories were out-of-date because WebExtensions was moved to a new product a few months ago.

This PR fixes that issue, and also adds a summary, introduction and icon.


helfi92:
I cleared the build cache and redeployed. Seems to work now :)

PR #251, opened 2018-10-14
Fix console warning in views/Project

### LINKED ISSUE #248: Fix warning in the console
screen shot 2018-10-08 at 11 07 00 pm

How to reproduce:
1. Run the localhost
2. Go to project that has information, e.g taskcluster
3. Inspect the console in browser

uu1t:
Fix #248.
Same as part of #79.

### DIFF
--- a/src/views/Project/index.jsx
+++ b/src/views/Project/index.jsx
@@ -369,7 +369,7 @@ export default class Project extends Component {
         {project.introduction && (
           <Card>
             <CardContent>
-              <Typography>
+              <Typography component="object">
                 <Markdown
                   source={
                     introductionOpen


eliperelman:
I'm curious why the component is `object`. I'm also not convinced that the other patch that used `object` is correct either.

fiennyangeln:
@eliperelman yup! I agree using object is misleading. I used it before after I read this https://material-ui.com/api/typography/ but I think I misinterpret the meaning 😅 . Hi @kojole are you interested in fixing that one too? I guess the component props means what will be the HTML tag like "div" will map into `<div ... />` and "span" will map into `<span .. />`

djmitche:
I'm closing this as stale.  It can be re-opened, or a new version created, of course!

PR #257, opened 2018-10-21
Sort projects dropdown (#255)

### LINKED ISSUE #255: Alphabetically sorted list of projects in dropdown
Right now the list of projects appear to be random (though they might not actually be). It would be much easier to browse if they were listed in the alphabetic order.
image

lashkari:
Hoping to resolve #255.

If compatibility < IE11 is needed, I can switch to using toLowerCase() instead of localeCompare()'s built-in options.

**Note:** I closed the previous pull request as I couldn't figure out how to re-run the failed checks after amending the commit. Apologies if this wasn't the right way to handle it.

### DIFF
--- a/src/components/TasksTable/index.jsx
+++ b/src/components/TasksTable/index.jsx
@@ -299,7 +299,15 @@ export default class TasksTable extends Component {
       project
     );
     const iconSize = 14;
-    const projects = [...new Set(items.map(item => item.project))];
+    const projects = [
+      ...new Set(
+        items
+          .map(item => item.project)
+          .sort((a, b) =>
+            a.localeCompare(b, undefined, { sensitivity: 'base' })
+          )
+      ),
+    ];
 
     return (
       <Fragment>


lashkari:
Just to be clear, the basic `localeCompare` (as used in `src/utils/sort.js`) is compatible back to IE6. It's the added options, including base letter insensitivity (e.g., `a = A = á = ä`), that's only supported in IE11 ([all other major browsers support it fine](https://caniuse.com/#feat=localecompare)).

An alternative solution to preserve compatibility with IE10 and older would be `a.toLowerCase().localeCompare(b.toLowerCase())`, but you lose the base letter insensitivity.

I'm guessing that your text-transform question was aimed more at other collaborators but I agree that, in general, it looks better. That said, it does make some entries look a bit odd when you'd expect camel-casing (e.g., "Github" instead of "GitHub", "Taskcluster" instead of "TaskCluster", etc.). Alternatively, you can make everything lowercase -- which still allows for a more consistent looking dropdown and somewhat eliminates the issue -- but doesn't really look _as_ good as capitalizing. Either way, you'd probably want to make sure that the project names are modified everywhere if you do this.

fiennyangeln:
@lashkari Good point! I don't think we need base letter support as of now looking at the current projects 👍 . 

Yeah that is my concern too. But as for me I realize that only Github repositories start with small letter and even when we capitalize it, it still have same meaning (we can go to the URL and look for the project in Github). That's why capitalizing came to my mind.. Also, for Github repositories, mostly are dash case, not camel case / pascal case.

However I feel like changing from dash case to pascal case is a bit too much (and we wont be able to look for the project in Github directly using the pascal case).

Let's hear what @helfi92 think about this and also regarding the sorting, shall we memoize it?

PR #268, opened 2018-11-12
Upgrade neutrino v.9

fiennyangeln:
TODO:
* [x] Setup jest properly
* [x] Make the icon that uses svg appearing (I tried variating the limit but it seems like when i print out the v9 config, the limit is still the default one). I also tried using svg-inline-loader
* [x] How to tidy up the webpack.config (it seems like it does not recognize `src` and `node_modules`)
* [x] How about contribute.json? (neutrino.on('build') is not a valid command, so I will look at ways to do it)


src/App/index.jsx
```diff
@@ -38,9 +38,16 @@ persistCache({
...
 })
 export default class App extends Component {
-  state = {
-    error: null,
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      error: null,
+    };
```
helfi92: Why not

```jsx
state = {
 error: null,
};
```

Is there a reason why we need the `constructor`?

.neutrinorc.js
```diff
@@ -37,16 +108,18 @@ module.exports = {
...
+      // ['ico', 'svg', 'img']
+      //   .forEach(rule => neutrino.config.module.rule(rule)
+      //     .use(images)
+      //     // .tap(options => ({ ...options, limit: 1 })))
+      //       .loader(require.resolve('@neutrinojs/image-loader'));
+      // neutrino.on('build', () => {
+      //   ['contribute.json'].forEach(file => {
+      //     fs.copyFileSync(file, join(__dirname, `build/${file}`));
+      //   })
+      // });
```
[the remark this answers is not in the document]
fiennyangeln: I keep it there to remind me to handle the contribute.json. Will replace it with the new syntax 👍

webpack.config.js
```diff
@@ -0,0 +1,8 @@
+const neutrino = require('neutrino');
+
+const config = neutrino().webpack();
+
+config.resolve.modules.push('src');
+config.resolve.modules.push('node_modules');
```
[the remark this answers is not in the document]
fiennyangeln: It does not seem to recognize the src and node_modules folder.
screen shot 2018-11-16 at 12 14 24 pm

fiennyangeln:
@helfi92 Hi! Made the suggested changes above + fixing jest not running properly by installing `babel-core@7.0.0-bridge.0` .. Now I guess what is left is to bring back the svg icons

.neutrinorc.js
```diff
@@ -12,23 +22,32 @@ module.exports = {
...
+      },
+      devServer: {
+        port: process.env.PORT,
+        historyApiFallback: { disableDotRule: true },
+      },
+      env: {
+        'GITHUB_PERSONAL_API_TOKEN' : '',
+        'BUGZILLA_ENDPOINT' : 'http://localhost:3090',
+        'NODE_ENV' : 'development',
+        'PORT': 5000,
```
helfi92: Remove `'` from the `env` object keys (e.g., `PORT` instead of `'PORT'`.

.neutrinorc.js
```diff
@@ -37,16 +56,13 @@ module.exports = {
...
-      neutrino.on('build', () => {
-        ['contribute.json'].forEach(file => {
-          fs.copyFileSync(file, join(__dirname, `build/${file}`));
-        })
-      });
+      // ['ico', 'svg', 'img']
+      //   .forEach(rule => neutrino.config.module.rule(rule)
+      //     .use(images)
+      //     // .tap(options => ({ ...options, limit: 1 })))
+      //       .loader(require.resolve('@neutrinojs/image-loader'));å
```
helfi92: Remove the comments.

.neutrinorc.js
```diff
@@ -37,16 +56,13 @@ module.exports = {
             .end()
           .use('gql-loader')
             .loader(require.resolve('graphql-tag/loader'));
+      neutrino.use(images, { limit: 1 });
```
helfi92: You can include that as an option to `@neutrinojs/react`:

```jsx
['@neutrinojs/react', {
  image: {
    limit: 1,
  },
  // ...
}]
```

package.json
```diff
@@ -42,14 +48,38 @@
...
+    "source-map-support": "^0.5.9",
     "typeface-roboto": "^0.0.54"
   },
   "devDependencies": {
-    "@neutrinojs/env": "^8.3.0",
-    "@neutrinojs/jest": "^8.3.0",
+    "@babel/plugin-proposal-class-properties": "^7.1.0",
+    "@babel/plugin-proposal-decorators": "^7.1.2",
+    "@mozilla-frontend-infra/react-lint": "^1.1.1",
+    "@neutrinojs/image-loader": "^8.3.0",
```
helfi92: You should be using the v9 version if this package.

fiennyangeln:
@helfi92 let me take a look, i think it could be because of me adding this `babel-core@7.0.0-bridge.0`

fiennyangeln:
@helfi92 hope this fixes the issue + I did the suggested changes. However I'm not sure if we need the port as the env since in my localhost, it seems to always choose 8080 / 8081 instead of 5000, the variable that I set.

package.json
```diff
@@ -42,14 +45,39 @@
...
-    "neutrino-preset-mozilla-frontend-infra": "^4.1.0"
+    "jest": "^23.6.0",
+    "js-yaml-loader": "^1.0.1",
+    "neutrino": "^9.0.0-beta.0",
+    "prettier": "^1.14.3",
+    "webpack": "^4.22.0",
+    "webpack-cli": "^3.1.2",
+    "webpack-dev-server": "^3.1.10"
+  },
+  "jest": {
```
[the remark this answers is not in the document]
fiennyangeln: Okay!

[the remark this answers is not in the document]
fiennyangeln: Test still run just fine

src/components/ProjectCard/index.jsx
```diff
@@ -53,20 +53,24 @@ export default class ProjectCard extends Component {
...
       }
 
-      const projectIcon = await import(/* webpackChunkName: "icon" */ `../../images/projectIcons/${
+      const projectIcon = (await import(/* webpackChunkName: "icon" */ `../../images/projectIcons/${
         project.fileName
-      }.svg`);
+      }.svg`)).default;
 
       this.setState({
         projectIcon: <img height="45" src={projectIcon} alt="Project Icon" />,
```
helfi92: Is there a reason why we are using ` ` to render the icon here and in the `catch` statement we are using a React element?

webpack.config.js
```diff
@@ -0,0 +1,5 @@
+const neutrino = require('neutrino');
+
+const config = neutrino().webpack();
+
+module.exports = config;
```
helfi92: Replace with:

```js
const neutrino = require('neutrino');

module.exports = neutrino().webpack();
```

src/components/ProjectCard/index.jsx
fiennyangeln: IIRC, in this case we are trying to get the project image from `src/images/projectIcons` folder if it exists. Otherwise if it throws error (the file does not exist), we use the WebIcon as the default icon. However, now it seems like none of the ProjectCard uses the default icon

src/components/Dashboard/index.jsx
```diff
@@ -109,12 +105,23 @@ export default class Dashboard extends Component {
...
   };
 
+  static defaultProps = {
+    classes: null,
+    withSidebar: true,
+    title: 'Codetribute',
+    header: null,
+  };
+
   static defaultProps = {
```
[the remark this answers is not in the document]
fiennyangeln: waah thanks for noticing!

.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 .Rhistory
 .env
 .eslintcache
+.DS_Store
```
helfi92: Add newline.

helfi92: `src/static/.DS_Store` seems to be the last one that needs to be removed.

fiennyangeln:
@helfi92 i realize the default padding is a bit different, e.g for the Sidebar in the homepage, shall we keep it as the new default value?

src/theme.jsx
```diff
@@ -36,16 +36,17 @@ const theme = createMuiTheme({
...
-    body1: Roboto400,
+    h1: Roboto300,
+    h2: Roboto400,
+    h3: Roboto400,
+    h4: Roboto400,
+    h5: Roboto400,
+    h6: Roboto500,
+    subtitle1: Roboto400,
+    body1: Roboto500,
+    body2: Roboto400,
```
helfi92: Now that we are using the next variants, we'll need to update the codebase. The new mapping is:

display4 => h1
display3 => h2
display2 => h3
display1 => h4
headline => h5
title => h6
subheading => subtitle1
body2 => body1
body1 (default) => body2 (default)

This means something like

```jsx
<Typography variant="display4" ... />
``` 

should become

```jsx
<Typography variant="h1" ... />
```

helfi92:
>  i realize the default padding is a bit different, e.g for the Sidebar in the homepage, shall we keep it as the new default value

Which padding are you talking about? Maybe you could show me with a screenshot or something :)

fiennyangeln:
What I do is I run the local version of this branch and compare it with the current live version. I realize that the `Skills` Sidebar entries look closer together in this local version, compared to the live version. After that I found the issue is that the default padding set by MaterialUI is different, the current version has `padding-top` and `padding-bottom` to be 11px, while the other has each padding to be 12px.
Below is the screenshot of each, if you look at the boxmodel section, it will show the padding difference.
screen shot 2019-01-10 at 5 34 25 pm
screen shot 2019-01-10 at 5 29 42 pm

fiennyangeln:
Thanks! Should we make the font size similar to the previous material version?

The last commit is to align the README guide that says 

> This will start a local development server on port 5000 (http://localhost:5000).

It is because, when I did not set the PORT in `.env` and just set it in the .neutrinorc.js it seems to point to any port available, e.g 8080, 8081.

src/theme.jsx
```diff
@@ -94,13 +94,20 @@ const theme = createMuiTheme({
...
       paperAnchorDockedLeft: {
         borderRight: 'none',
       },
     },
+    MuiListItem: {
+      root: {
+        paddingTop: '12px',
+        paddingBottom: '12px',
+      },
+    },
```
helfi92: Why are we adding `MuiListItem`?

README.md
```diff
@@ -44,6 +44,7 @@ To get started with local development, create a file in the root of the repo nam
 ```bash
 GITHUB_PERSONAL_API_TOKEN=<your_github_api_token>
 BUGZILLA_ENDPOINT=http://localhost:3090
+PORT=5000
```
helfi92: Do we use `PORT` outside of `.neutrinorc.js`? If not, then we can remove `PORT` from here and have the value be hard-coded. Also remove `NODE_ENV : 'development'` from `.neutrinorc.js` since webpack does that for us when we do `"dotenv webpack-dev-server -- --mode development",`.

```diff
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -28,14 +28,12 @@ module.exports = {
         favicon: './src/static/favicon.png',
       },
       devServer: {
-        port: process.env.PORT,
+        port: 5000,
         historyApiFallback: { disableDotRule: true },
       },
       env: {
         GITHUB_PERSONAL_API_TOKEN : '',
         BUGZILLA_ENDPOINT : 'http://localhost:3090',
-        NODE_ENV : 'development',
-        PORT: 5000,
       },
       babel: {
         plugins: [
```

src/theme.jsx
fiennyangeln: The padding is different in the Sidebar, Material v2 has 11px while the v1 has 12px.

README.md
fiennyangeln: Yes, we dont use it anywhere else apart from the .neutrinorc.js.. Sure!

fiennyangeln:
Yes, the font weight is the same, however I feel like the font-size is a bit different. If I open the netlify preview and the current live side by side, there is some difference, but the difference is small. Should we handle that or leave it to match the current version's default?

src/theme.jsx
helfi92: Ah I see. It should still be okay to remove it. The UX still look nice :)

fiennyangeln:
Yeah, I agree! @helfi92

helfi92:
I can merge once Travis finishes.

fiennyangeln:
Thanks for the guidance @helfi92 !! 😆
    On Tuesday, 22 January 2019, 10:05:18 PM SGT, Hassan Ali <notifications@github.com> wrote:  
 
 
@helfi92 approved this pull request.

 🎉 Thank you @fiennyangeln, you did a fantastic job!

djmitche:
That was one of the longer PR's I've been involved with :)

Nice work!

Issue #274, opened 2018-12-18
Switch to using React.lazy


helfi92:
Let's remove `react-loadable` and use `React.lazy` instead which was introduced in React v16.6.

https://reactjs.org/docs/code-splitting.html#reactlazy

For inspiration, you can look at how it's done in another repository https://github.com/taskcluster/taskcluster-web/pull/206.

armenzg:
`react-lodable` was breaking my HMR support:
https://github.com/mozilla/bugzilla-dashboard/commit/626af8c429e6188eda3801e624f5a8161a51a3ab

djmitche:
I think it is!  Please get started based on the above information.  Although we are not as "available" as usual right now due to work holidays, we are happy to answer questions.

fiennyangeln:
Let's close this issue ? @helfi92

PR #284, opened 2019-02-13
Use React.lazy

fiennyangeln:
Solves #274


src/utils/loadable.jsx
```diff
@@ -12,29 +11,14 @@ import Spinner from '../components/Spinner';
...
+    );
   }
 }
-
-export default loader =>
-  Loadable({
-    loader,
-    loading: Loading,
-    timeout: 10000,
-  });
```
helfi92: We should be able to safely remove ` src/utils/loadable.jsx` assuming it is not used anymore.

fiennyangeln: I think currently it is used in the Suspense part of `src/App/index.jsx`

helfi92: Oh, in that case we should use <Spinner /> directly instead of <Loading />. By the way, The Spinner component has a `loading` prop which would center the svg. That might remove the need of doing something like `textAlign: 'center'`.

fiennyangeln: Agree! let me try 😄

src/App/index.jsx
```diff
@@ -81,7 +85,8 @@ export default class App extends Component {
                     path={path}
                     exact={exact}
                     render={({ staticContext, ...renderProps }) => (
-                      <Suspense fallback={<Loading />}>
+                      <Suspense
+                        fallback={<Spinner className={classes.spinner} />}>
```
helfi92: Are we supposed to provide the `loading` prop to cover for the old style `textAlign: 'center',`?

```js
fallback={<Spinner loading ... />}>
```

fiennyangeln: I think it is not needed. When I test the spinner, it looks centered.
screen shot 2019-02-14 at 2 13 13 pm

PR #322, opened 2019-08-11
Add infinite loading table

A1igator:
There are a few important notes with this pull request:

- [react-virtualized](https://github.com/bvaughn/react-virtualized/) was used as raect-window does not have a straightforward way to have a table yet. The implementation was based on [mui virtualized table example](https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/tables/ReactVirtualizedTable.js).
- The table appears to require a height, which can not be based upon number of items. (1000px right now)
- Bugzilla GraphQL seems to be acting weirdly, as it is saying there are no pages left, when there are pages left, no matter the page size.
- There doesn't seem to be a way to give rows keys, as they had before.
- Since now it renders a column as opposed to a row, I am not sure how to include the "there doesn't seem to be any items to show".
- The propTypes have not been done yet. I will do them when I know everything else with the pull request is fine.


A1igator:
>I think we can use the AutoSizer component to have the height and width managed depending on how much space is left but I see that you are already using that so I'm not sure I understand the question.

I was basically talking about the double scrollbar issue which you addressed right under your comment and the WindowScroller component worked very well, so thank you!
>How can I reproduce this issue?

change BUGZILLA_PAGE_SIZE in constants.js to a smaller number like 5. console.log bugzillaData in Project view. You will see the pageInfo object from the GraphQL query not making sense, as it says there aren't anymore pages when there clearly is, since you can change BUGZILLA_PAGE_SIZE back to 100 and get more elements. (I tested this in the Firefox Desktop Frontend project. Not sure if different projects behave differently)

>The Table component of react-virtualized exposes a noRowsRenderer which we could probably use to render the empty state.

Sort of fixed. The code is quite ugly. I'll come up with a better way soon. Thanks.

I also realized there is another bug as the column sizes no longer adjust when the elements overflow.(ie in the rust project the tags don't fit) I will need to research a bit on how to fix that as it comes with the way the column widths are coded.

A1igator:
Based on this [issue](https://github.com/bvaughn/react-virtualized/issues/468), I'm not sure if it is possible to fix the Column sizing problem. I welcome any suggestions you may have.

src/views/Languages/index.jsx
```diff
@@ -99,8 +100,13 @@ const tagReposMapping = repositories =>
...
         },
-        paging: {
-          ...BUGZILLA_PAGING_OPTIONS,
+        pagingGood: {
+          page: 0,
+          pageSize: BUGZILLA_PAGE_SIZE,
+        },
+        pagingMen: {
+          page: 0,
+          pageSize: BUGZILLA_PAGE_SIZE,
```
helfi92: I think the reason why paging is not working is because `BUGZILLA_PAGING_OPTIONS` was removed here. When I run the master branch and set the page size to 5, `hasNextPage` is set to true:

Screen Shot 2019-08-19 at 10 11 39 AM

A1igator: But bugzilla paging option is just this:
`BUGZILLA_PAGING_OPTIONS = {
  page: BUGZILLA_PAGE_NUMBER,
  pageSize: BUGZILLA_PAGE_SIZE,
};`
where BUGZILLA_PAGE_SIZE is defaulted to 100 and BUGZILLA_PAGE_NUMBER is defaulted to 0. It is true that in that spot I could change it to BUGZILLA_PAGING_OPTIONS, but I don't think that will solve the issue. Did you try it on Firefox Desktop Frontend project? (I ask because you commented on the Languages part of the code, as opposed to Projects)

helfi92: > Did you try it on Firefox Desktop Frontend project? (I ask because you commented on the Languages part of the code, as opposed to Projects)

Ah, I didn't notice where I commented. I actually tried in on `localhost:5000/projects/taskcluster`. I just looked at the network tab to see the request being sent and noticed that `page` is set to `null` instead of 0. Perhaps that's the reason?

Screen Shot 2019-08-19 at 10 56 59 AM

A1igator: I did some looking and that null occurs because the response before it was
 ```
pageInfo {
  "hasNextPage": false,
  "nextPage": null,
}
```
The first request is 0 and not null though and the response is the same as above.

Comment (A1igator) [reply src/views/Languages/index.jsx]: image
image

Here are comparison screenshots for the column problem. an HTML5 table appears to adjust cell width based on the content while react virtualized does not.

Comment (fiennyangeln) [reply src/views/Languages/index.jsx]: Thanks for picking this up. I think I found the reason for the below.

>  Bugzilla GraphQL seems to be acting weirdly, as it is saying there are no pages left, when there are pages left, no matter the page size.

### TLDR: 

`this.load` function only fetch product with components. When you inspect the page and set the paging to be low, it seems that on the first query (product without component) you get the next page (1). However, you call `this.load` with the next page (1), the other variables (product, component) is different.

### Long thoughts:
#### Background
For bugzilla, the data stored in yaml is like this
```
products:
 - Firefox
 - Toolkit
 - Cloud Services: ['Firefox Sync: UI']
```

This means that product can have component, e.g in above case 'Firefox Sync: UI' or skip the component name.

When we query, we first only query bug for the product with no component, e.g in above case ['Firefox', 'Toolkit'] then in the [this.load function](https://github.com/mozilla-frontend-infra/codetribute/blob/master/src/views/Project/index.jsx#L241) we fetch bug for each product with component.

The reason for that is that product and component relationship is not one to many. There could be one bug with same component, yet belongs to different product. Thus, we agreed to not combine the query.

Thus, the query done initially (in the `@graphql`) is for fetching product with no component. And [this.load function](https://github.com/mozilla-frontend-infra/codetribute/blob/master/src/views/Project/index.jsx#L241) is meant for fetching product with component.

#### Proof of my theory
Remove the  `- Cloud Services: ['Firefox Sync: UI']` line from `ff.yaml` and set the paging size to be 5. It should work.

#### Possible fixes
_I'm throwing some ideas_

1. Make it such that when `loadNextPage` is called, it also fetches product with no component. 
```
const productWithNoComponentList = project.products.filter(product => typeof product === 'string');
    await this.fetchBugzilla(productWithNoComponentList, undefined);
```
2. Separate the paging information for each product-component query parameter. I think this is important, not only for Bugzilla, but also for Github (where we use different query for different label). Some query last longer than other, imagine how if the paging cursor have been updated to 4 by some query result of param A, yet other query of param B still haven't fetch the bug for paging 3.

Comment (A1igator) [reply src/views/Languages/index.jsx]: I'll get on doing fix 2 as it seems to be the better option. Thank you very much!

Comment (A1igator) [reply src/views/Languages/index.jsx]: actually I suppose I have to do both fixes if I'm reading this right.

**Comment by A1igator** [comment]
This works but I am not sure if it is the best way of coding it due to global variables. The only other place I could think of would be in the the this.props.bugzilla or this.props.github, but I couldn't figure out how to setup dotProp with a dictionary.

**Comment by helfi92** [comment]
@fiennyangeln Did you want to take a look at this by any chance? :-) Thanks!

File: src/components/DataTable/index.jsx [hunk 2]
```diff
@@ -62,48 +56,112 @@ class DataTable extends Component {
...
+                        )}>
+                        <em style={{ color: 'black' }}>
+                          No items for this page.
+                        </em>
+                      </TableCell>
+                    )}>
+                    {columns.map(({ dataKey, ...other }, index) => (
+                      <Column
+                        key={dataKey}
+                        headerRenderer={headerProps =>
```
fiennyangeln: As the suggestion above to remove `this.headerRenderer` we can make it into something like below

```suggestion
                        headerRenderer={({ label }) => (
                          <TableCell
                            className={classes.flexContainer}
                            variant="head"
                            style={{ height: headerHeight }}>
                            <TableSortLabel
                              id={label}
                              active={label === sortByHeader}
                              direction={sortDirection || 'desc'}
                              onClick={this.handleHeaderClick}>
                              {label}
                            </TableSortLabel>
                          </TableCell>
                        )}
```

src/views/Project/index.jsx
```diff
@@ -190,19 +326,23 @@ export default class Project extends Component {
         const moreGoodFirstNodes = fetchMoreResult.goodFirst.edges;
         const moreMentoredNodes = fetchMoreResult.mentored.edges;
 
-        // return previousResult;
         if (!moreGoodFirstNodes.length && !moreMentoredNodes) {
           return previousResult;
         }
 
+        pagingGood[JSON.stringify({ products, components })] =
```
fiennyangeln: I think we can use only products here as I think it should be unique in the list.

src/views/Languages/index.jsx
```diff
@@ -215,69 +187,45 @@ export default class Languages extends Component {
...
-      [];
-    const goodFirstBugs =
-      (bugzillaData &&
-        bugzillaData.goodFirst &&
-        uniqBy(
+    const bugzillaData = this.props.bugzilla;
+    let goodFirstBugs = [];
+    let mentoredBugs = [];
+
+    if (bugzillaData) {
```
fiennyangeln: can this be 
```suggestion
    const goodFirstBugs =
      (bugzillaData &&
        bugzillaData.goodFirst &&
        uniqBy(
         .....
```

src/views/Languages/index.jsx
```diff
@@ -291,18 +239,21 @@ export default class Languages extends Component {
...
-        bugzillaData.mentored &&
-        uniqBy(
+          'summary.title'
+        );
+
+        ({ hasNextPage } = bugzillaData.goodFirst.pageInfo);
+      }
+
+      if (bugzillaData.mentored) {
+        mentoredBugs = uniqBy(
```
fiennyangeln: can this be as below? thus the if is no longer needed. Also, can we do this in the render method. Thus, we don't need to keep the items state.
```suggestion
    const mentoredBugs =
      (bugzillaData &&
        bugzillaData.mentored &&
        uniqBy(
```

src/views/Languages/index.jsx
```diff
@@ -215,69 +187,45 @@ export default class Languages extends Component {
...
+    await this.load();
+    const githubData = this.props.github;
+    let issues = [];
+    let hasNextPage = false;
 
-        return comments[0].text;
-      } catch (error) {
-        this.setState({ error });
-      }
+    if (githubData && githubData.search) {
```
fiennyangeln: can we make it simpler & move it to the render method?
```suggestion
    let issues = this.props.github && this.props.github.search && uniqBy( 
```

src/views/Languages/index.jsx
```diff
@@ -111,10 +117,9 @@ const tagReposMapping = repositories =>
 )
 export default class Languages extends Component {
   state = {
-    error: null,
-    // we need a way to know when github loading is done.
-    // github results require calling fetchMore possibly many times.
-    githubLoading: false,
+    hasNextPage: true,
+    isNextPageLoading: false,
+    items: [],
```
fiennyangeln: Can we remove this state? Based on what I see, it seems that we can obtain this from props. Thus by using state, there seem to be 2 sources of truth, state & props. I think it is better if we do as before, obtaining the items by combining `githubData` and `bugzillaData` in the render method. What do you think?

src/views/Project/index.jsx
```diff
@@ -123,8 +131,123 @@ const tagReposMapping = repositories =>
...
+    await this.setState({ isNextPageLoading: true });
+    await this.load();
+    const githubData = this.props.github;
+    const bugzillaData = this.props.bugzilla;
+    let issues = [];
+    let hasNextPage = false;
+    let goodFirstBugs = [];
+    let mentoredBugs = [];
+
+    if (githubData && githubData.search) {
```
fiennyangeln: Same as above. Use `&&` instead of if

src/views/Project/index.jsx
```diff
@@ -123,8 +131,123 @@ const tagReposMapping = repositories =>
...
+      );
+
+      Object.keys(pagingGithub).forEach(x => {
+        if (pagingGithub[x].hasNextPage) {
+          hasNextPage = true;
+        }
+      });
+    }
+
+    if (bugzillaData) {
```
fiennyangeln: Same as above. Use `&&` is prefered compared to if.

src/views/Project/index.jsx
```diff
@@ -163,6 +286,14 @@ export default class Project extends Component {
     const {
       bugzilla: { fetchMore },
     } = this.props;
+    const pageGood =
+      JSON.stringify({ products, components }) in pagingGood
+        ? pagingGood[JSON.stringify({ products, components })].nextPage
```
fiennyangeln: Can we try using only products, I believe it should be unique already

src/views/Project/index.jsx
```diff
@@ -229,10 +372,12 @@ export default class Project extends Component {
           return previousResult;
         }
 
+        pagingGithub[searchQuery] = fetchMoreResult.search.pageInfo;
+
         return dotProp.set(
           previousResult,
           'search.nodes',
-          moreNodes.concat(previousResult.search.nodes)
+          previousResult.search.nodes.concat(moreNodes)
```
fiennyangeln: Any reason for this change? Since this PR is quite large in size, let's strive to minimize the number of lines changed as much as we can.

src/views/bugs.graphql
```diff
@@ -1,13 +1,17 @@
-query Bugs($goodFirst: BugSearch!, $mentored: BugSearch!, $paging: Paging) {
-  goodFirst: bugs(search: $goodFirst, paging: $paging) {
+query Bugs($goodFirst: BugSearch!, $mentored: BugSearch!, $pagingGood: Paging!, $pagingMen: Paging!) {
```
fiennyangeln: Can we rename `pagingGood` to `goodFirstPaging` possibly and `pagingMen` to `mentoredPaging`? I think of the new name after the `$goodFirst` & `mentored` variables.

src/views/Project/index.jsx
```diff
@@ -42,6 +43,9 @@ const tagReposMapping = repositories =>
       ...mappings,
     };
   }, {});
+const pagingGithub = {};
+const pagingGood = {};
+const pagingMen = {};
```
fiennyangeln: @helfi92 any advise on where to track all the paging cursor variables?

helfi92: Maybe something like

```jsx
const pageCursors = {
  github: {},
  bzGoodFirst: {},
  bzMentored: {},
};
```

A1igator:
Thanks for all the helpful suggestions :) I believe I have applied all of them to the Project view. If you think Project view is good now, I can go ahead and apply them to Languages view.

src/views/Project/index.jsx
```diff
@@ -337,35 +384,62 @@ export default class Project extends Component {
...
+
+    if (!hasNextPage) {
+      Object.keys(pageCursors.bzGoodFirst).forEach(x => {
+        hasNextPage = !hasNextPage
+          ? pageCursors.bzGoodFirst[x].hasNextPage
+          : true;
+      });
+    }
+
+    if (!hasNextPage) {
```
fiennyangeln: What do you think about replacing all the if block of `hasNextPage` to
```suggestion
    const hasNextPage =
      Object.keys(pageCursors.github).some(
        x => pageCursors.github[x].hasNextPage
      ) ||
      Object.keys(pageCursors.bzGoodFirst).some(
        x => pageCursors.bzGoodFirst[x].hasNextPage
      ) ||
      Object.keys(pageCursors.bzMentored).some(
        x => pageCursors.bzMentored[x].hasNextPage
      );
```

src/views/Project/index.jsx
```diff
@@ -268,33 +304,42 @@ export default class Project extends Component {
...
   };
 
   render() {
-    const { classes } = this.props;
+    const { isNextPageLoading, error } = this.state;
     const githubData = this.props.github;
-    const { loading, error } = this.state;
+    const bugzillaData = this.props.bugzilla;
     const project = projects[this.props.match.params.project];
+    let hasNextPage = false;
```
fiennyangeln: If we apply the below, we can get rid of this line
```suggestion
```

fiennyangeln (dismissed):
I haven't looked into styling! But generally it looks good to me, just a little nit! Feel free to make the changes in the `Languages` too.

A1igator:
In regards to languages, just to make sure, I will need to make a fetchBugzilla function, because it appears at the moment, you only fetch bugzilla at the start.

src/views/Project/index.jsx
```diff
@@ -271,14 +271,14 @@ export default class Project extends Component {
     });
   };
 
-  load = async () => {
+  load = () => {
```
fiennyangeln: Hi, can you elaborate the reason you remove the `async` from the function?

fiennyangeln:
Regarding the fetchBugzilla function, yes I think it might be necessary. Previously we did not need it because we can use one request to fetch it as we don't need to split based on product / component.

src/views/Project/index.jsx
A1igator: Yes, when it was called by loadNextPage, Promise.all for fetchGithub had issues with it being async and awaited, causing it to load the same query from the same cursor twice, as opposed to loading a query and loading the one after with the new cursor. After 2 tries, loadNextPage just gave up. Due to this github data only displayed the first query. I got rid of the await on the Promise.all and that fixed it. I got rid of await on everything else in that function too after, even though fetchBugzilla wasn't having the same problems at the moment, I thought it might in the future. When I got rid of all the awaits, the async was no longer needed, and I believe the function works perfectly now.

A1igator:
I've almost fully setup Languages, but there appears to be a bug with "withSidebar" prop of "Dashboard" component causing only 20 items to be rendered, even though it leaves enough space for all the items.

A1igator:
I fixed the bug I believe. Unless you guys want me to figure out a way to increase cell size based on content (the only way I see now is to hardcode a value based on text size for each column), everything is done, except propTypes.

fiennyangeln:
Hi! Thanks for fixing this! Sorry for my late reply, I was on a business trip last week. Feel free to add the propTypes, overall looks good to me. I'm trying to find way to display this nicely on mobile browser.
### Current master
Screen Shot 2019-10-14 at 12 14 27 PM

### This branch
Screen Shot 2019-10-14 at 12 14 22 PM

A1igator:
> Hi! Thanks for fixing this! Sorry for my late reply, I was on a business trip last week. Feel free to add the propTypes, overall looks good to me. I'm trying to find way to display this nicely on mobile browser.
> ### Current master
> 
> Screen Shot 2019-10-14 at 12 14 27 PM
> ### This branch
> 
> Screen Shot 2019-10-14 at 12 14 22 PM

That comes back to this [issue](https://github.com/bvaughn/react-virtualized/issues/468). Unless someone else has already solved this problem before, you will be required to simulate the width sizing of an HTML table with adjusting the widths of each column in TasksTable when it's passed into DataTable.

src/views/Project/index.jsx
```diff
@@ -238,12 +271,42 @@ export default class Project extends Component {
...
+        : []
+    );
+    const productWithNoComponentList = project.products
+      ? project.products.filter(product => typeof product === 'string')
+      : [];
 
-    await Promise.all(
+    if (
+      !(
+        Object.entries(productWithComponentList).some(
```
fiennyangeln: nits. if you only need the key of the mapping, you can use `Object.keys`
```suggestion
        Object.keys(productWithComponentList).some(products =>
```

src/views/Project/index.jsx
```diff
@@ -238,12 +271,42 @@ export default class Project extends Component {
...
+      project.products
+        ? project.products.filter(product => typeof product !== 'string')
+        : []
+    );
+    const productWithNoComponentList = project.products
+      ? project.products.filter(product => typeof product === 'string')
+      : [];
 
-    await Promise.all(
+    if (
```
fiennyangeln: Does this `if` block mean to reset the cursor information? Is this used when user moves to different Project page and come back? If that's the case, can we set it to be empty whenever user enters this page, e.g using componentDidMount? Otherwise, feel free to add comments on this.

A1igator: You are correct. I fixed it. At first I just reused the same logic as Languages, where you can't use component did mount because the table is all that changes and not the whole component in Languages.

A1igator:
I somewhat fixed mobile. I'm not sure if it's good enough.

helfi92:
I'm planning to look at this soon. Apologies for the delay!

A1igator:
@helfi92 Sorry for the ping. I just wanted to make sure this is still on your radar.

helfi92:
> @helfi92 Sorry for the ping. I just wanted to make sure this is still on your radar.

Looking now :)

A1igator:
@helfi92 Sorry I was busy for a while. I believe number 2 is fixed. Regarding number 1, I am not actually limiting the width of the table at all (because master isn't either), and it should grow with screen size. I can only reproduce what you are seeing via zooming out (I do not have a big monitor) and is easily fixed by refreshing. I have added an event-listener to change width on resize so a refresh is no longer required.

A1igator:
Hi @helfi92 @fiennyangeln, 

I know you must be busy, especially now, due to corona. I also know I took way too long to implement the changes requested. I would really love to get this pull request merged, and I do have some free time now myself as school is over if you'd like anymore changes.

helfi92:
Hey @A1igator. Sorry for the wait. I will try to make some time to review this. This feature will make a lot of people happy :)

A1igator:
@helfi92 Odd. It is working for me. Assuming you are on the right fork, try Firefox Desktop Frontend section? Scroll to the bottom and wait a second and it should load another set of entries. Some of the language like "Rust" will also load entries.

A1igator:
@helfi92 I just tried it out on a new machine and it's working just as I expected. Is there something you are expecting it's not doing?

helfi92:
Ah, it looks like there was something wrong with my bugzilla api key. I created a new one and now it works. Will review now.

src/views/Project/index.jsx
```diff
@@ -190,7 +218,11 @@ export default class Project extends Component {
         const moreGoodFirstNodes = fetchMoreResult.goodFirst.edges;
         const moreMentoredNodes = fetchMoreResult.mentored.edges;
 
-        // return previousResult;
+        pageCursors.bzGoodFirst[JSON.stringify(products)] =
+          fetchMoreResult.goodFirst.pageInfo;
+        pageCursors.bzMentored[JSON.stringify(products)] =
+          fetchMoreResult.mentored.pageInfo;
```
helfi92, line 224: Is there a way to do this without shoving the full list in the pageCursors object (i.e., `JSON.stringify(products)`)?

src/components/TasksTable/index.jsx
```diff
@@ -142,16 +146,43 @@ const assignments = Object.values(ASSIGNEE);
...
+     * Whether next page is loading.
+     */
+    isNextPageLoading: bool,
+
+    /**
+     * A function to execute when next page is to be loaded.
+     * No arguments are needed.
+     * This can be used to get more data from the APIs.
+     */
+    loadNextPage: func,
```
helfi92, line 175: Let's call this `onNextPageLoad`

src/components/TasksTable/index.jsx
```diff
@@ -363,20 +430,60 @@ export default class TasksTable extends Component {
...
         </div>
-        <div className={classes.tableWrapper}>
-          <DataTable
-            items={data}
-            renderRow={item => (
-              <TableRow tabIndex={-1} key={item.summary}>
+        <DataTable
+          columns={[
+            {
+              width: Math.max((windowWidth / 2000) * 180, 100),
```
helfi92, line 436: Can you explain how you came with these numbers. I want to understand them :)

src/views/Languages/index.jsx
```diff
@@ -216,7 +204,108 @@ export default class Languages extends Component {
       })
     );
 
-    this.setState({ githubLoading: false });
+    if (bugzillaData) await this.fetchBugzilla(language);
+
+    this.setState({ isNextPageLoading: false });
```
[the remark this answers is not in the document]
A1igator: It shows up at the top of the page, so since you scroll down, you won't see it unless you scroll back up. I do realize this isn't ideal but I wasn't sure if I should change the spinner code to overlap the table and stick to the top of the view.

src/components/TasksTable/index.jsx
A1igator: Mostly by just inspect elementing the old table, and try to find numbers that match the cell proportions.

src/views/Project/index.jsx
```diff
@@ -140,6 +149,12 @@ export default class Project extends Component {
     }
   }
 
+  componentDidMount() {
+    pageCursors.github = {};
+    pageCursors.bzGoodFirst = {};
+    pageCursors.bzMentored = {};
+  }
```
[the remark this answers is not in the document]
A1igator: The reason why those are global variables is because updateQuery overwrites the "this" keyword to refer to itself, so I cannot access state in that function.

src/views/Project/index.jsx
A1igator: I'm not sure. Maybe I could make a unique hash from the products, then shove that?

src/views/Project/index.jsx
```diff
@@ -337,35 +387,51 @@ export default class Project extends Component {
...
+    const hasNextPage =
+      Object.keys(pageCursors.github).some(
+        x => pageCursors.github[x].hasNextPage
+      ) ||
+      Object.keys(pageCursors.bzGoodFirst).some(
+        x => pageCursors.bzGoodFirst[x].hasNextPage
+      ) ||
+      Object.keys(pageCursors.bzMentored).some(
+        x => pageCursors.bzMentored[x].hasNextPage
+      );
```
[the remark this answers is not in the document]
A1igator: Fixed!

src/components/TasksTable/index.jsx
A1igator: Fixed!

src/views/Project/index.jsx
helfi92: That makes sense to me.

A1igator: I got rid of the JSON.stringify by joining each product into one string, which will help performance. I don't actually think a hash will help here, because based on my knowledge, it is not possible to make a unique hash shorter than the string.

src/views/Languages/index.jsx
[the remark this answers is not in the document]
helfi92: Ah, ok I understand now. For a better user experience, I think we should show the spinner in the bottom when the list is loading. Have you had a look at https://github.com/bvaughn/react-virtualized/blob/master/docs/creatingAnInfiniteLoadingList.md? It looks like we can integrate this with react-virtualized.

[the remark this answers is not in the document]
A1igator: Didn't use that method but I believe I achieved the same effect.

A1igator:
Hey @helfi92 is there anything else?

djmitche:
@helfi92 should we merge this or close it as stale?

djmitche:
Thanks @helfi92 - I'll rebase and take a look.

@fiennyangeln I see you requested changes here -- would you like to look again, or have you forgotten everything about this? :)

djmitche:
This also needs a rebase, and is well beyond my capacity to review.  I would be fine with clicking "merge" and seeing what happens, given a rebase.

djmitche:
Unfortunately, it looks like this has fallen behind in the months since it was completed.  I'm going to close the PR, but this would make a good starting point for someone more familiar with how infinite scrolling works who wants to finish the issue.

PR #341, opened 2019-10-07
Treeherder to support Github issues

armenzg:
I want to start using Github to get contributors since most people are comfortable around it than Bugzilla.


armenzg:
Thank you for the fast turnaround!
It's working!
image

PR #384, opened 2020-02-26
Update outdated information in DevTools description

juliandescottes:
Starting with the bare minimum.

- removed mention related to team size (no strong reason, but it's inaccurate and seems irrelevant)
- updated IRC link to Matrix
- removed link to the debugger repository

We still aggregate profiler's good first bugs, so I didn't remove the mentions related to GitHub. I think it would make sense to split the profiler in its own "product" but it's not my call to make. 

cc @nchevobbe @violasong 

Do you feel like we should also rewrite the content a bit?


violasong:
Looks good! We can think about rewriting the content after I've made my updates to the main docs.

nchevobbe:
(looked good for me too, thanks Julian)

PR #396, opened 2020-04-08
Add Go to the sidebar

djmitche:
I see that the JavaIcon for example is defined as an SVG inline.  Where did that come from, and could I find something similar for Go?

Also, in testing this out it looks like I'm getting 401's from GitHub for all of the `https://codetribute.mozilla.org/languages/foo` URLs, including this one.  So I'm not 100% sure this works :)


djmitche:
Hm, they're not working on my development site but do seem to be working in production, so maybe it's an API key issue.  Anyway, hopefully this is simple enough!

djmitche:
Updated with a Go icon

PR #410, opened 2020-08-14
Group DevTools components

nchevobbe:
I also took the opportunity to remove the mention of the public Slack instance that we plan to close soon

I don't know why, but the `What's New` item does not show up. Maybe something is wrong because of the quote?

### DIFF
--- a/src/data/devtools.yaml
+++ b/src/data/devtools.yaml
@@ -41,8 +41,75 @@ introduction: |
 
   Some bugs have an assigned mentor, and that person will usually be the one to reply.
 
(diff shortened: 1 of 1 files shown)


juliandescottes:
Regarding `What's New`, it might just be that the component has no mentored bug? DOM doesn't show up either and similarly has no mentored bugs.

helfi92:
> Regarding `What's New`, it might just be that the component has no mentored bug? DOM doesn't show up either and similarly has no mentored bugs.

That would make sense.

djmitche:
Oh good, I was worried that was an escaping issue with the `'` in the name!

PR #417, opened 2020-09-24
automerge renovate PRs for all but major updates

djmitche:
I'll also add renovate-bot with write access to the repo

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,5 +9,11 @@
     "helfi92",
     "fiennyangeln"
   ],
-  "pinVersions": false
+  "pinVersions": false,
+  "packageRules": [
+    {
+      "updateTypes": ["minor", "patch", "lockFileMaintenance"],
+      "automerge": true
+    }
+  ]
 }


djmitche:
Actually I think / hope it doesn't need more access than what it gets from the GitHub app integration.

helfi92:
Do we really want this? We don't actively watch the site to make sure it's not breaking. The past couple of times where the site broke it's because an external user bothered creating a new issue. What do you think?

djmitche:
Yeah, that's a good point.  However, the last time was my fault, unwisely merging a major version bump.  This PR would only automerge minor/patch bumps.

I see this repo as a good example of a project that is "maintained" but doesn't have dedicated support.  I think keeping dependencies up-to-date in a low-touch fashion is going to be important for such projects.  In particular, I noticed this project has a lot of unmerged renovate PRs, which suggests that keeping up with renovate is requiring too much human time.

I don't know if there are any "easy" ways to improve automated testing, either, especially since the tricky bits are probably the graphql queries.

Maybe the risk of breakage is a manageable one?

helfi92:
Sounds good. We can give it a try.

helfi92 (approved):

djmitche:
I also removed the requirement for reviews before merging, and restricted pushes to master to just admins + renovate, all of which are [require](https://github.com/renovatebot/config-help/issues/885) to make this work.

PR #419, opened 2020-09-24
Revert "fix(deps): update dependency react-apollo to v3"

djmitche:
This reverts commit c9350b46373bbed1413d8f82e0c6cfccd3d7aef2.  The upgrade to react-apollo v3 is really a switch to @react/client, which is more like a rewrite, it seems.


djmitche:
merging since this has the site broken..

PR #494, opened 2021-03-31
Add Data @ Mozilla

wlach:
Used Taskcluster as a template. Going to get a bit of feedback from my peers before asking for this be formally merged / reviewed.


wlach:
@mdboom Any thoughts on this? This might make a nice funnel for getting more people involved in our projects. Here's a quick preview of what it looks like: https://deploy-preview-494--mozilla-codetribute.netlify.app/projects/data

src/data/data.yaml
```diff
@@ -0,0 +1,43 @@
...
+  Individual data projects may have project-specific channels or forums for asking questions and getting advice.
+  We are also available on [matrix.io](https://chat.mozilla.org/#/room/#telemetry:mozilla.org) in the `#telemetry` channel in the mozilla.org community.
+
+products:
+  - "Data Platform and Tools"
+
+repositories:
+  - mozilla/glean: "good first issue"
+  - mozilla/glean.js: "good first issue"
+  - mozilla/glean-dictionary: "good first issue"
```
wlach, line 43: If we want this to supersede "Telemetry Dashboards", we should add the relevant sections here:

```suggestion
  - mozilla/glean-dictionary: "good first issue"
  - mozilla/telemetry-dashboard: ['mentored', 'good first issue']
```

mdboom:
I think this is great.  LGTM, pending @chutten's thoughts on whether to combine with Telemetry Dashboards or not.

src/data/data.yaml
chutten: I'm happy with this superseding "Telemetry Dashboards"

@djmitche I notice that there's a ghost "project" for Telemetry as well (`Toolkit::Telemetry` on BMO, part of Firefox Desktop). Should we umbrella that here, too? Or would that be a little odd.

djmitche: I'm just the person responsible for clicking merge here -- y'all know the projects better than me :D

brizental: nit: Both the Glean SDK and Glean.js actually use a [good first **bug**] tag. Not sure that makes a difference here. Also, when we start getting a bit of traction on Glean.js I am thinking about having a [help wanted] tag for bugs that might not be great for first constributions, but welcome contributor help. I guess it would be no harm to already include that tag here.

wlach: @brizental oops, good catch! We actually want to index the bugzilla components for glean + glean.js, not github. I'll fix this. It seems like codetribute has its own heuristics for finding good first bugs and mentored bugs based on the bugzilla conventions, so long as you align with those I think you should be good:

https://github.com/mozilla-frontend-infra/codetribute/blob/main/src/views/Languages/index.jsx#L98

wlach:
Ok, I think this is ready to go, except I'm not seeing Glean.js in: 

https://deploy-preview-494--mozilla-codetribute.netlify.app/projects/data

This is odd because I definitely see some good first bugs under the component here: https://bugzilla.mozilla.org/buglist.cgi?f1=status_whiteboard&v1=%5Bgood%20first%20bug%5D&o1=substring&resolution=---&query_format=advanced&f2=component&v2=Glean.js&o2=equals&classification=Client%20Software&classification=Developer%20Infrastructure&classification=Components&classification=Server%20Software&classification=Other&list_id=15657481

@djmitche you wouldn't know offhand why this isn't working, do you?

djmitche:
Nothing jumps out at me, no.  I can see the GraphQL request for that data, and it's the same request that looks for `Glean: SDK` bugs, which do appear.

My guess is that it has to do with the `.` in the component name.

Should we get this landed and then work to debug that issue?  The GraphQL bugzilla backend is currently un-owned and I've never even cloned it, so fixing that could be a bit of a slog.

wlach:
> Nothing jumps out at me, no. I can see the GraphQL request for that data, and it's the same request that looks for `Glean: SDK` bugs, which do appear.
> 
> My guess is that it has to do with the `.` in the component name.

Yeah, I had a similar intuition. It probably needs to be url-escaped or something.

> Should we get this landed and then work to debug that issue? The GraphQL bugzilla backend is currently un-owned and I've never even cloned it, so fixing that could be a bit of a slog.

Makes sense to me, I can take a very quick look next week.

djmitche:
Ah, I see, the query is:

```
  "goodFirst": {
			"components": [
				"Glean.js",
				"Glean: SDK"
			],
			"keywords": [
				"good-first-bug"
			],
			"order": "changeddate DESC",
			"products": [
				"Data Platform and Tools"
			],
			"statuses": [
				"NEW",
				"UNCONFIRMED",
				"ASSIGNED",
				"REOPENED"
			]
		},
```

but the Glean.js bugs seem to have the whiteboard field set, not the keyword.  And, adding that keyword makes them appear in codetribute.

djmitche:
> Codetribute will read the `products` field from the yaml file and will extract bugs with keyword `good-first-bug`. Here are some ways to display bugs on the site.                                                                                                                                                                                                                             

So, this behavior is currently hard-coded -- I think the fix is to move those whiteboard entries to keywords.

wlach:
Thanks @djmitche that makes sense. Last I remember, keyword searches are faster than whiteboard tags.

PR #512, opened 2021-04-20
Bug:Fix For Focus On Select Filters

Patil2099:


djmitche:
Can you explain a bit about how this helps?

Patil2099:
Focus is required for keyboard accessibility. We could change the style a little but it is okay for now and the expected behavior.

djmitche:
Thanks for looking at this!  Sometimes a PR or issue ends up not changing any code, but it's still a win because we've learned something :)

PR #558, opened 2022-04-09
filter out PRs from issue queries

### LINKED ISSUE #557: Rust project is not opened
Hi :)
When I enter the Rust project in codetribute, the Rust project is not opened.

Screenshot from 2022-04-07 01-04-33

URL) https://codetribute.mozilla.org/projects/rust

djmitche:
I'm not sure if there's a more appropriate way to filter these out in the graphql query, but this works!

Fixes #557.

### DIFF
--- a/src/views/Project/index.jsx
+++ b/src/views/Project/index.jsx
@@ -336,7 +336,11 @@ class Project extends Component {
       (githubData &&
         githubData.search &&
         uniqBy(
-          githubData.search.nodes.map((issue) => ({
+          githubData.search.nodes
+          // filter out PRs, which appear in issue queries but aren't
+          // issues (and don't have a repository property, for example)
+          .filter(issue_or_pr => issue_or_pr.__typename === "Issue")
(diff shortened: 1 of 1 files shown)


jcristau:
cc @mhoye

djmitche:
@petemoore does your team still own codetribute?

petemoore (commented):
I don't know who owns codetribute these days. :-/

I can rubber stamp, if you've been able to test locally etc, or if you could add a test that runs in the CI, that would be even better, just to be sure this doesn't break anything.

petemoore:
I think this will get deployed to production on merge to main (via netlify), is that right?

djmitche:
I believe that's correct -- or at least was correct.

I don't think codetribute has tests.  It's been a while, but I vaguely recall testing this.  At any rate, if it has an issue it's easy to back out.

lotas (approved):

lotas:
Yay! I finally can merge those :)

PR #581, opened 2022-08-18
Separate our developer infrastructure to a separate section.

Standard8:
Some of the lint/source docs and other items just got [moved into a different project in bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1784867).

This PR is to add a new section for the new parts and make the build system section a little clearer.

### DIFF
--- a/src/data/ff-build.yaml
+++ b/src/data/ff-build.yaml
@@ -1,9 +1,7 @@
-name: Firefox - Development Environment
-summary: The harnesses, code analysis, and linting tools surrounding the building of Firefox.
+name: Firefox - Development Build System Environment
+summary: The harnesses tools surrounding the building of Firefox.
 icon: package-variant
 introduction: |
   The development environment (aka build system) supports the building of Firefox.
-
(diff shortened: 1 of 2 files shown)


src/data/ff-build.yaml
```diff
@@ -1,9 +1,7 @@
-name: Firefox - Development Environment
-summary: The harnesses, code analysis, and linting tools surrounding the building of Firefox.
+name: Firefox - Development Build System Environment
```
djmitche, line 1: That seems like a lot of general words.  Is this a team name, or is there something more succinct or specific?

djmitche (commented):
Looks good to me, but I am not in a position to merge or deploy.

src/data/ff-build.yaml
Standard8: Yeah... I kinda want to make it after devtools & co, but I could drop the `Environment` part.

lotas:
Thanks! I can merge it now :)

PR #602, opened 2023-03-16
Update Fenix instructions and include the "GeckoView" Bugzilla product

cpeterso:
Fenix bug tracking has moved from GitHub to Bugzilla, which uses different keywords and terminology.


lotas (approved):
Thank you!

PR #614, opened 2024-02-13
Add Kitsune/SuMo

smithellis:
We'd like to add Kitsune/SuMo ( support.mozilla.org ) to Codetribute.


lotas (approved):
Thanks for adding this!

PR #622, opened 2024-06-25
Delete Sentence Collector

MichaelKohler:
The Common Voice Sentence Collector was integrated into the Common Voice platform some time ago. Therefore the separate repository was archived.


lotas:
Thanks!

